PR #257, opened 2025-01-03
add cookie consent prompt for all users

DavidHoerster:
Add a cookie consent prompt for all users when first navigating to the site.


merill (approved):
Looks good.

PR #262, opened 2025-01-16
fix test result text to include mention of CA policies

DavidHoerster:
Provide a more descriptive text about the strong auth methods test result.


merill (approved):
LGTM!

PR #306, opened 2025-05-21
Add Test-Assessment-21777 to verify App Instance Property Lock config…

alexandair:
Add test code and detailed explanation for Test-Assessment-21777 to verify App Instance Property Lock configuration for multitenant applications.


merill (approved):
Looks good

PR #309, opened 2025-05-22
Add Test-Assessment-21786 to verify Token Protection policies

alexandair:
Add test code and detailed explanation for Test-Assessment-21786 to verify Token Protection policies.


merill (approved):
Looks good!

PR #323, opened 2025-05-29
Add fix for tenant detection

JasSuri:
Add the capability for differentiated tests based on tenant type.
The tenant type is obtained from /organization endpoint and can be either AAD (Workforce tenant) or CIAM (External tenant).

The tests are moved into a JSON file from which we can mark whether the test is applicable to AAD, CIAM or both.
This can be expanded with other attributes, such as "template": "CISA", to enhance the runtime of any particular tenant specific test set.


merill (approved):
Looks good. Made a few refactors

PR #329, opened 2025-06-01
Add Test-Assessment-22128 to check for guest users in privileged roles

alexandair:
Add `Test-Assessment-22128` to check for guest users in privileged roles and provide detailed reporting.


merill (approved):
Looks good!

PR #334, opened 2025-06-04
Add Test-Assessment-21818 to validate activation alerts

alexandair:
Add Test-Assessment-21818 to validate activation alerts for privileged roles and improve notification checks.


merill (approved):
Looks good.

PR #350, opened 2025-06-07
Add Global Administrator role test (21817) to validate approval workflow

alexandair:
Add Global Administrator role test (21817) to validate approval workflow requirements and report results.


merill (approved):
Looks good.

PR #351, opened 2025-06-09
Integrate initial EEID tests

JasSuri:
Added two tests #24069 and #24070, which are EEID specific checks.


src/powershell/private/tests/Test-Assessment.24070.ps1
```diff
@@ -0,0 +1,94 @@
...
+    Write-ZtProgress -Activity $activity
+
+    $caps = Invoke-ZtGraphRequest -RelativeUri 'identity/conditionalAccess/policies' -ApiVersion beta
+
+    $mfaPolicies = $caps | Where-Object {`
+            $_.grantControls.builtInControls -contains "mfa"}
+
+    $protectedApps = $mfaPolicies.Conditions.Applications.IncludeApplications
+
+    $userFLows = Get-MgIdentityAuthenticationEventFlow
```
merill, line 24: Can we use Invoke-ZtGraphRequest instead of the cmdlet (we want to avoid having dependencies on various cmdlets, also Invoke-ZtGraphRequest caches the api call so it will be faster when it's used in other tests.

src/powershell/private/tests/Test-Assessment.24070.ps1
```diff
@@ -0,0 +1,94 @@
...
+    SELECT * FROM SignIn WHERE appId IN ('$guidList')
+"@
+
+        $results = Invoke-DatabaseQuery -Database $Database -Sql $sql
+        $signInsPerUnprotectedApp = $results.appId | Group-Object | Sort-Object Count -Descending | Select-Object @{Name="AppId"; Expression={$_.Name}}, Count
+
+
+        $appNames = @()  # Create an empty array to store results
+        foreach ($appId in $unprotectedApps) {
+            $app = Get-MgApplicationByAppId -AppId $appId
```
merill, line 50: You might want to query the database since we already have all the apps in the tenant in the db.

src/powershell/private/tests/Test-Assessment.24070.ps1
```diff
@@ -0,0 +1,94 @@
...
+
+    if ($passed) {
+        $testResultMarkdown = "Tenant is configured to require multi-factor authentication for all user flows protectecting consumer apps.`n`n%TestResult%"
+    }
+    elseif (!$passed) {
+        $testResultMarkdown = "Tenant is not configured to require multi-factor authentication for all user flows protectecting consumer apps.`n`n"
+        $testResultMarkdown += "Found $totalUnprotectedApps consumer apps that are not requiring multi-factor authentication.`n`n%TestResult%"
+        $guidList = $unprotectedApps -join "','"
+        $sql = @"
+    SELECT * FROM SignIn WHERE appId IN ('$guidList')
```
merill, line 41: We might want to avoid querying the signins table from the database. We only download 1hr worth of logs because it can get quite large in very large tenants. So the actual sign in data might only be for a few hours or days and might not be 30 days of data. 

It would have been neat if the signin graph api supported count then we could have pulled it directly from graph by app id but that's not supported. I think you might want to just remove the sign in count. You could alternatively show lastSignedIn date if your intention is to show active/inactive apps.

src/powershell/private/tests/Test-Assessment.24070.ps1
```diff
@@ -0,0 +1,94 @@
...
+    Add-ZtTestResultDetail -TestId '24070' -Title 'Consumer facing apps are protected with MFA'`
+        -UserImpact Medium -Risk Medium -ImplementationCost Low `
+        -AppliesTo Identity -Tag User, Credential `
+        -Status $passed -Result $testResultMarkdown -GraphObjectType ConditionalAccess -GraphObjects $unprotectedApps
+}
+
+function Build-table($Apps) {
+    $mdInfo = ""
+    foreach ($item in $Apps) {
+        $mdInfo += "| $($item.DisplayName) | $($item.AppId) | $($item.SignIns)|  `n"
```
merill, line 91: We try to provide a deep link to the app in the Entra portal. See some of the other tests. 

Makes it easy for users to go directly into the app to review and make changes.

merill (changes requested):
I've added bunch of comments on some of the changes.

merill:
Closing as this code is now out of date as we've moved away from .json test meta

PR #352, opened 2025-06-09
Refactor test functions to use array syntax for database query results

alexandair:
Refactor test functions to use array syntax for database query results and format date outputs consistently.

Just wanted to format date outputs consistently using the `Get-FormattedDate` function and use splatting for the `Add-ZtTestResultDetail` command.

But found a bug in tests: 21772, 21773, and 21992.

```powershell
    $resultsApp = Invoke-DatabaseQuery -Database $Database -Sql $sqlApp
    $resultsSP = Invoke-DatabaseQuery -Database $Database -Sql $sqlSP

    $passed = ($resultsApp.Count -eq 0) -and ($resultsSP.Count -eq 0)
```

Sometimes, `Invoke-DatabaseQuery` returns an array of hash tables. In that case, $resultsApp.Count and $resultsSP.Count correctly return a number of hash tables.
However, if the result contains only one hash table, `Count` property returns a number of hash table's key-value pairs. That leads to incorrect test results.

**Fix**: Wrapping the call in @(...) forces an array. Regardless of how many results come back, you always get an array:
- If zero records, you get an empty array (@()), not $null.
- If one record, you get a single-element array containing that object.
- If multiple records, you get an array of those objects (same as unwrapped in many cases, but consistently an array).

```powershell
    $resultsApp = @(Invoke-DatabaseQuery -Database $Database -Sql $sqlApp)
    $resultsSP = @(Invoke-DatabaseQuery -Database $Database -Sql $sqlSP)

    $passed = ($resultsApp.Count -eq 0) -and ($resultsSP.Count -eq 0)
```

### DIFF
--- a/src/powershell/private/tests/Test-Assessment.21772.ps1
+++ b/src/powershell/private/tests/Test-Assessment.21772.ps1
@@ -1,4 +1,3 @@
-
 <#
 .SYNOPSIS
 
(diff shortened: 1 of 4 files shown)


alexandair:
@merill I've removed "array wrapping" workaround, so that you can accept changes with `Get-FormattedDate` and splatting.

PR #354, opened 2025-06-11
Add Test-Assessment-21865 to validate trusted network locations

alexandair:
Add Test-Assessment-21865 to validate trusted network locations for improved risk detection and Conditional Access policies.


merill (approved):
Merging this version for now. Will add in the updates as they come.

PR #362, opened 2025-06-14
Add Test-Assessment-21886 to validate application provisioning

alexandair:
Add `Test-Assessment-21886` to validate application provisioning for Microsoft Entra, including detailed reporting on unconfigured applications.

This test has the same issue when `Invoke-ZtRequestGraph` returns a single object with `value` property equals to `{}`.
It also has performance issues with API calls.


merill (approved):
Looks good

PR #368, opened 2025-06-21
Add Test-Assessment-21868 to check for guest user ownership of apps

alexandair:
Add `Test-Assessment-21868` to check for guest user ownership of applications and service principals, including detailed reporting and remediation actions.

This test is quite slow due to the multiple API calls. It needs 5 minutes to finish for our test tenant.
I've written a version with jobs and JSON batching, but it uses `Invoke-MgGraphRequest`.
`Invoke-ZtGraphRequest` doesn't support batching. 
Let's discuss the desired approach.


merill (approved):
Looks good

PR #381, opened 2025-06-28
Add Test-Assessment-21858 to identify inactive guest users

alexandair:
Add `Test-Assessment-21858` to identify inactive guest users and generate detailed reports.


merill (approved):
Looks good

PR #382, opened 2025-06-28
Add Test-Assessment-21874 to test external collaboration domain checks

alexandair:
Add Test-Assessment-21874 to enhance external collaboration domain checks.

The test title in the spec and TestMeta.json are not the same. I've used one from the spec. It was overwritten.


merill (approved):
Looks good

PR #383, opened 2025-06-29
Add Test-Assessment-21811 to check for password expiration settings

alexandair:
Add `Test-Assessment-21811` to check for password expiration settings and generate detailed reports on misconfigured domains and users.


merill (approved):
Looks good

PR #393, opened 2025-07-09
Add interactive configuration wizard for Zero Trust Assessment

alexandair:
I've added `-Interactive` and `-ConfigurationFile` parameters to `Invoke-ZtAssessment`.
All interactivity is in new `New-ZtInteractiveConfig.ps1` file.
TUI is based on `PwshSpectreConsole` module. For now, that module is not added to .psd1 file, but you can install it manually. You can also test how `Invoke-ZtAssessment` works before you install the module.

@merill, you wanted to add support for excluding emergency accounts from tests. Could you provide a spec for that?


merill (approved):
This is a great start. Thanks a lot for adding this @alexandair

PR #402, opened 2025-08-09
Update app-permissions.md

DeanGross144:
add information to explain what will and won't be checked by the script


merill (approved):
Thanks @DeanGross144 this helps clarify!

PR #421, opened 2025-09-08
Implement private Get-PolicyAssignmentTarget function 

alexandair:
Implement `Get-PolicyAssignmentTarget` function and remove duplicates from assessment tests


merill (approved):
Looks good.

PR #425, opened 2025-09-09
Adding PR Validation

FriedrichWeinmann:
Adding lots of tests to ensure code quality

Current Requirements:

- All public commands must have full help
- All script files must not contain syntax errors
- All script files must not contain undesired commands
- Functions may not be duplicated across the project
- Each script file under public & private may only contain a single top-level function

None of these requirements are _enforced_ until the check has been set as required for a given branch.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1,0 +1,15 @@
+﻿on: [pull_request]
+
+jobs:
+  validate:
+
+    runs-on: windows-latest
+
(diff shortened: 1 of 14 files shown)


FriedrichWeinmann:
Looks like my module bootstrapper is also failing right now, making the Github Action not run the tests at all. Will investigate and fix later today

merill (approved):
Looks good!

PR #442, opened 2025-09-12
24564 - Add assessment test for Local Users and Groups policy

alexandair:
24564 - Add assessment test for Local Users and Groups policy.


merill (approved):
LGTM

PR #445, opened 2025-09-15
test 24573 - Windows Security Baseline

gaelcolas:
To be merged after #444

This is the one I need clarification on when Intent not showing sometimes...

Result:
image


merill (approved):
Looks good!

PR #451, opened 2025-09-16
24568 - Add assessment test for macOS SSO policy configuration

alexandair:
24568 - Add assessment test for macOS SSO policy configuration and assignment.

It contains a helper function that filters policies based on specific setting configurations and their expected values.
If the helper function proves useful in other tests as well, we will make it a private function in the module.


merill (approved):
LGTM

PR #457, opened 2025-09-18
Add Get-FilteredPoliciesBySetting private function

alexandair:
Add `Get-FilteredPoliciesBySetting` private function to filter policies based on specific settings and expected values.


merill (approved):
LGTM

PR #461, opened 2025-09-18
Feature 21847

sandeepjha000:
added new test


merill (commented):
@sandeepjha000 fyi we want to avoid including https://graph.microsoft.com in the url

This is because the code can run in Gov cloud, and other environments where the graph url could be different. I'll fix it for this one.

merill:
@sandeepjha000 Please accept the CLA license agreement too. Tx

KManoj0000:
@microsoft-github-policy-service agree [company="Perennial Sys"]

PR #465, opened 2025-09-19
Tab Expansion Update and Dev Convenience

FriedrichWeinmann:
A small, yet fun update:

> Tab Expansion

The Tab Expansion limit of 10 tests offered is fine as a default, and limiting the number of options offered was an inspired design, especially when looking at ultimately hundreds of tests. That said, the specific number of 10 is a good default, but depending on screen & console size, more options might be useful.

While 10 is still the default, users are now able to change that number!

```powershell
# Only for the current console
Set-PSFConfig -Module ZeroTrustAssessment -Name 'TabExpansion.TestLimit' -Value 20

# Remember it for next time
Set-PSFConfig -Module ZeroTrustAssessment -Name 'TabExpansion.TestLimit' -Value 20 -PassThru | Register-PSFConfig
```

> Dev Experience

When working on tests, running the assessment is all fine and well. But when you want to test a new version, you have to import the module again (or test it isolated from the module), which means discarding all the current module state, cached values, etc.
Doable yes, but costs more time to iterate.

Well ... you _had_ to import the module again.

Introducing the new test reload option:

```powershell
Get-ZtTest -Reload
```

This will reload all the test script files and update functions, aliases and variables as needed.

Note: This parameter is hidden from the regular user and will _not_ show up in tab completion to the user, until it is the only remaining possible parameter to complete.
While having a Get-* command do this seems a bit off-focus, I wanted to avoid adding a public command, visible to regular users, that basically is only useful to devs. I also wanted to avoid adding another tool outside of the module for the purpose and this seemed a nice balance of benefit, cost and impact.


merill (approved):
LGTM

PR #492, opened 2025-09-28
Test 21823 - Guest self-service sign up disabled

gaelcolas:
Result:
image

and fail:

image


merill (changes requested):
Can we make `Guest self-service sign up via user flow` in the message clickable so it takes the user to the admin portal page where they can configure this setting? tx

gaelcolas:
Added the link to the Azure blade.
Ready for review

PR #496, opened 2025-09-29
24570 - Add test assessment for Entra Connect Sync configuration

alexandair:
24570 - Add test assessment for Entra Connect Sync configuration.


merill (changes requested):
Did you have trouble with the -Filter parameter of Invoke-ZtGraphRequest?

Also can we include a space between the icon and text: '❌Yes'

Tx

alexandair:
@merill 
I think I never used `-Filter`. But, I didn't have a problem with filter. I had the problem escaping `$select` inside `$expand`.
Solution was to use `EscapeDataString()` method.

How would you like me to rewrite it?

PR #499, opened 2025-09-29
Feature 21879

KManoj0000:


merill:
@KManoj0000  The demo tenant has multiple assignments for external users but is not being shown and the test is being marked as a Pass incorrectly.

Merill 2025-10-15 at 12 25 51@2x

merill (approved):
Merging to test

PR #638, opened 2025-11-25
Updated Connect-ztAssessment module to include service principal login

naveedious:
Allow both the MgGraph and azure connection to use clientid / service principal. This is so assessments can be done in an automated fashion without interactive logins.


naveedious:
@microsoft-github-policy-service agree

mlhickey:
Variation - this started as the use of a local/custom clientId for auth instead of the public MG endpoint.  Simple enough to add teh clientId as a param for Connect-ZtAssessment and as a param for Connect-MgGraph:

        # The client ID to use for authentication. If not specified, the default client ID will be used.
        [string]$ClientId
  
        if ($ClientId)
        {
            $params['ClientId'] = $ClientId
        }

merill:
Good call @mlhickey 

We've added support for a custom ClientId. It will come out in the next release.

See https://github.com/microsoft/zerotrustassessment/blob/main/src/powershell/public/Connect-ZtAssessment.ps1

PR #654, opened 2025-11-28
Bugfix/issue 579 - Test 21835 (Emergency accounts) failing

### LINKED ISSUE #579: Test 21835 (Emergency accounts) failing
Possibly same root cause as #266, but creating a separate issue for tracking since those still are two distinct tests. 

I've run the assessment against two tenants (one of them cloud-only), both failing at this test. The result looks similar:
Summary:
- Total permanent Global Administrators: 3
- Cloud-only GAs with phishing-resistant auth: 0
- Emergency access accounts (excluded from all CA): 0
- Enabled Conditional Access policies: 9

There seems there might be a logic bug checking for phishing-resistant auth:
```
foreach ($method in $authMethods) {
    $methodType = $method.'@odata.type'
    $authMethodTypes += $methodType

    # If any method is not FIDO2 or Certificate, mark as not phishing-resistant only
    if ($methodType -ne '#microsoft.graph.fido2AuthenticationMethod' -and
        $methodType -ne '#microsoft.graph.x509CertificateAuthenticationMethod') {
        $hasOnlyPhishingResistant = $false
    }
}
```

The problem is that in addition to a YubiKey with FIDO2 I have password as an auth method. The YubiKey is the preferred default and the CA policies should prevent this from being valid for logging in as well. (I have one targeting the break glass account requiring phish-resistant MFA.) But as far as I know passwords cannot be properly removed from a user object, and it's not possible to remove/disable it under the "Authentication Methods" blade in the Entra Portal either. (Or an API visible in the Graph Explorer.) In other words the if condition is not going to work as intended.

Since this will give a zero result there are no emergency candidates to verify in the next step.

Or is just me missing something obvious here?

sandeepjha000:
Test 21835 was failing because the current logic checks if users have only phishing-resistant authentication methods (FIDO2 and/or Certificate). However, since passwords are always present in the authentication methods array, the condition $hasOnlyPhishingResistant will always become $false, resulting in zero emergency account candidates.

Solution:
The logic is updated to check at least one phishing-resistant authentication method, rather than checking if they have only phishing-resistant methods. This aligns with the security objective - if CA policies enforce phishing-resistant MFA, the presence of a password doesn't matter.

fixes #579

### DIFF
--- a/src/powershell/tests/Test-Assessment.21835.ps1
+++ b/src/powershell/tests/Test-Assessment.21835.ps1
@@ -68,7 +68,7 @@ WHERE vr.roleDefinitionId = '62e90394-69f5-4237-9190-012177145e10'
 
     foreach ($user in $permanentGAUsers) {
         # Only process cloud-only accounts (onPremisesSyncEnabled is null or false)
-        if ($null -eq $user.onPremisesSyncEnabled -or $user.onPremisesSyncEnabled -eq $false) {
+        if ($user.onPremisesSyncEnabled -ne $true) {
             Write-PSFMessage "Checking auth methods for cloud-only user: $($user.userPrincipalName)" -Level Verbose
 
             # Use Get-ZtUserAuthenticationMethod helper to get authentication methods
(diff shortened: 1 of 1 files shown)


merill (approved):
LGTM

PR #666, opened 2025-12-02
Feature-25405 - Intelligent Local Access is enabled and configured

Manoj-Kesana:
Intelligent Local Access is enabled and configured.

### DIFF
--- a/src/powershell/tests/Test-Assessment.25405.ps1
+++ b/src/powershell/tests/Test-Assessment.25405.ps1
@@ -69,15 +69,15 @@ function Test-Assessment-25405 {
         $mdInfo += "Found $networkCount private network(s) configured for Intelligent Local Access.`n`n"
 
         $formatTemplate = @'
-| Name | Id |
+| Network name | Id |
 | :--- | :--- |
 {0}
 
 '@
         foreach ($network in ($privateNetworks | Sort-Object name)) {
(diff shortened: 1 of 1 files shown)


alexandair (changes requested):
The logic flow is incorrect:
- Gets networks
- If $null, skips as NotSupported (Why? We never do that. We test if it exists or not.)
- Counts networks and checks if > 0

Problem: The test will never fail.

Line 66: TestId should be a string

Please, organize code in regions like it's done in test 25392.

How did you test the behavior when there is a private network?

src/powershell/tests/Test-Assessment.25405.ps1
```diff
@@ -0,0 +1,99 @@
...
+    #region Data Collection
+    Write-PSFMessage '🟦 Start' -Tag Test -Level VeryVerbose
+
+    $activity = 'Checking Intelligent Local Access configuration'
+    Write-ZtProgress -Activity $activity -Status 'Getting private networks'
+
+    # Query private networks from Global Secure Access
+    $privateNetworks = Invoke-ZtGraphRequest -RelativeUri 'networkaccess/privateNetworks' -ApiVersion beta
+
+    # Initialize test variables
```
alexandair: Move variables into Assessment Logic region.

PR #670, opened 2025-12-03
Bugfix/issue-661: Test 21783 is reporting custom Entra ID roles as not being enforced

### LINKED ISSUE #661: Privileged Microsoft Entra built-in roles are targeted with Conditional Access policies to enforce phishing-resistant methods
Still reporting custom Entra ID Roles as not being enforced. CA Policy blade doesn't allow you to select custom roles on policy.

Custom role needs to not be checked and flagged as failed.

sandeepjha000:
Issue in test 21783:
Conditional Access policies cannot target custom roles but test is validating against custom privileged roles as well.

Solution:
Only consider built in privileged roles for testing if they are targeted with Conditional Access policies to enforce phishing-resistant methods

fixes #661


merill (approved):
LGTM

PR #672, opened 2025-12-04
Added check to exclude Terms of Use CA policies - Test 21824

### LINKED ISSUE #592: Guests don't have long lived sign-in sessions
Terms of Use CA policy should be excluded as it has its own sign-in frequency.

praneeth-0000:
Issue: Current test 21824 doesn't filter Terms of Use CA policies, which have their own sign in frequency.

Resolution: Exclude Terms of Use CA policies from the test.

fixes #592


alexandair (approved):
LGTM

merill (approved):

merill (approved):

PR #676, opened 2025-12-05
feature-25370: User source IP addresses are preserved for Conditional Access and risk detection

sandeepjha000:
Initial commit for network pillar test 25370
image

image


alexandair (changes requested):
Please, organize code in regions like in test 25392.
TestId should be a string.
Don't return early (lines 31-40), we are also testing when the check fails.

alexandair (commented):
@sandeepjha000  Regions should have the same names as in test 25392.

alexandair (approved):
@sandeepjha000  Thank you for addressing requested changes. Add @tdetzner as a reviewer.

PR #692, opened 2025-12-10
Feature-25399 (Network) - Private DNS is configured for internal name resolution  

sandeepjha000:
Test 1:
image

Test 2:
image

Test 3:
image

### DIFF
--- a/src/powershell/tests/Test-Assessment.25399.ps1
+++ b/src/powershell/tests/Test-Assessment.25399.ps1
@@ -31,19 +31,21 @@ function Test-Assessment-25399 {
     $activity = 'Checking Private DNS configuration for Quick Access (Entra Private Access)'
     Write-ZtProgress -Activity $activity -Status 'Querying Quick Access application'
 
-    # Q1: Find Quick Access application
+    # Query 1: Find Quick Access application
     $quickAccessApp = Invoke-ZtGraphRequest -RelativeUri "applications" -Filter "tags/any(c:c eq 'NetworkAccessQuickAccessApplication')" -ApiVersion beta
+    #endregion Data Collection
 
+    #region Assessment Logic
     # Initialize test variables
     $testResultMarkdown = ''
     $passed = $false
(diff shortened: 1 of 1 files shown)


src/powershell/tests/Test-Assessment.25399.ps1
```diff
@@ -0,0 +1,120 @@
...
+    # Initialize test variables
+    $testResultMarkdown = ''
+    $passed = $false
+    $appDnsResolutionEnabled = $false
+    $appDnsSuffixes = @()
+    $appHasValidSegments = $false
+    #endregion Data Collection
+
+    #region Assessment Logic
+    # Assessment Logic 1: Check if Quick Access application exists
```
alexandair: Assessment Logic is the name of the region, not the query. Correct other occurrences as well.

src/powershell/tests/Test-Assessment.25399.ps1
```diff
@@ -0,0 +1,132 @@
...
+
+    # Check if Quick Access application exists
+    if (-not $quickAccessApp -or $quickAccessApp.Count -eq 0) {
+        $testResultMarkdown = "❌ No Quick Access application found with 'NetworkAccessQuickAccessApplication' tag."
+        $passed = $false
+    }
+    else {
+        # Get the Quick Access application
+        $app = $quickAccessApp
+        $appId = $app.id
```
alexandair: Does $quickAccessApp always contain a single object?
If it can contain multiple objects, $app.id will return multiple IDs and the next Invoke-ZtGraphRequest command will fail.

alexandair (commented):
@sandeepjha000 Please, address new feedback.

src/powershell/tests/Test-Assessment.25399.ps1
sandeepjha000: @SagarSathe confirmed that we can have only one Quick Access App.

alexandair (approved):
LGTM

PR #695, opened 2025-12-11
Fixing throttling & dev experience update

FriedrichWeinmann:
This PR wraps up some loose ends from the two main PRs that parallelized tests & exports.
Changes:

> Updated FunctionsToExport for improved Git handling

Essentially, with the old style - comma-separated in the same line - two PRs that add a new public function would always conflict.
Changing it to one line per function exported prevents that and works just as well.

> Fixed Throttling parameter in Invoke-ZtAssessment

Previously, the `-ExportThrottleLimit` parameter was unused. This was mostly due to me pushing the parameters in one go during PR one, originally planning to do both parallelization refactors in one go.
This proved impractical, but I did not want to deal with more conflicts than needed, so I left that for later, when both PRs were merged. Now.

> Added Format XML for Export Statistics

The output of `Get-ZtExportStatistics` now looks better, without losing any of the information contained.
Also removed a redundant XML setting that had no effect anyways.

> Added Export Statistics to the Support Package

When generating a Support Package for sending in debug data for the ZeroTrustAssessment, both test & export statistics are now included:

```powershell
Expand-Archive -Path .\powershell_support_pack_2025_12_11-14_30_35.zip -DestinationPath .
$data = Import-PSFClixml -Path .\powershell_support_pack_2025_12_11-14_30_35.cliDat
$data._ZTA_TestStatistics
$data._ZTA_ExportStatistics
```

### DIFF
--- a/src/powershell/ZeroTrustAssessment.format.ps1xml
+++ b/src/powershell/ZeroTrustAssessment.format.ps1xml
@@ -1,6 +1,41 @@
 ﻿<?xml version="1.0" encoding="utf-16"?>
 <Configuration>
     <ViewDefinitions>
+        <!-- ZeroTrustAssessment.ExportStatistics -->
+        <View>
+            <Name>ZeroTrustAssessment.ExportStatistics</Name>
+            <ViewSelectedBy>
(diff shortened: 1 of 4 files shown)


merill (approved):
LGTM

PR #702, opened 2025-12-12
adding Network Test - 25371

komalp2025:
PR for spec - https://github.com/microsoft/ztspecs/pull/91

issue https://github.com/microsoft/ztspecs/issues/26


alexandair (changes requested):
@komalp2025 Please, address my feedback.

src/powershell/tests/Test-Assessment.25371.ps1
```diff
@@ -0,0 +1,224 @@
...
+function Get-UserDisplayNameById {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory = $true)]
+        [string[]]$UserIds,
+        [Parameter(Mandatory = $true)]
+        $Database
+    )
+    $UserResult = @()
+    if($UserIds -eq 'All'){return 'All'}
```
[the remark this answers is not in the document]
komalp2025: if conditional access policy applicable to all users, than it will return "All", to encounter that have written this check.

image

src/powershell/tests/Test-Assessment.25371.ps1
```diff
@@ -0,0 +1,225 @@
...
+        if ($caePolicies -and $caePolicies.Count -gt 0) {
+            foreach ($policy in $caePolicies) {
+                $targetUsersDisplayNames = @()
+                $appCondition = $policy.conditions.applications
+                # Primary check: Check if policy targets All applications
+                $targetsAllApps = $appCondition.includeApplications -contains "All"
+
+                # Extract target users and groups
+                $targetUsers = @($policy.conditions.users.includeUsers)
+                if($targetUsers){$targetUsersDisplayNames = Get-UserDisplayNameById -UserIds $targetUsers -Database $Database}
```
merill: I know the spec states display user name and group. But we don't do this for CA policies in general and it can cause performance issues and also make the list long. Please remove the displaying of users and groups, the user should be able to click through and view the details.

merill (dismissed):

src/powershell/tests/Test-Assessment.25371.ps1
```diff
@@ -0,0 +1,159 @@
...
+
+    # Q2: Check traffic forwarding profiles status (Prerequisite)
+    # Determine which GSA traffic forwarding profiles are active
+    Write-ZtProgress -Activity $activity -Status 'Getting traffic forwarding profiles'
+    $forwardingProfiles = Invoke-ZtGraphRequest -RelativeUri 'networkAccess/forwardingProfiles' -ApiVersion beta
+
+    # Q3: Check for Conditional Access policies that disable CAE for GSA traffic
+    # Query enabled Conditional Access policies to identify any that explicitly disable CAE
+    Write-ZtProgress -Activity $activity -Status 'Checking Conditional Access policies'
+    $caePolicies = Get-ZtConditionalAccessPolicy | Where-Object { $_.state -eq 'enabled' }
```
alexandair: You didn't resolve Merill's comment:
Please call graph without filter and do the filter locally. This is what we do in all our other tests and need to remain consistent. In fact we have Get-ZtConditionalAccessPolicy and that's what should be used.

src/powershell/tests/Test-Assessment.25371.ps1
```diff
@@ -0,0 +1,159 @@
...
+        Pillar = 'Network',
+        RiskLevel = 'High',
+        SfiPillar = 'Protect networks',
+        TenantType = ('Workforce'),
+        TestId = 25371,
+        Title = 'Network access is validated in real-time through Universal Continuous Access Evaluation',
+        UserImpact = 'Low'
+    )]
+    [CmdletBinding()]
+    param()
```
alexandair: The current implementation suffers from the "N+1 query problem," where it executes a separate database round-trip for every single user ID. If a policy targets 20 users, that's 20 queries. If you have 50 policies, that could be 1000 queries, significantly slowing down the assessment.

Here is the optimized version using a single IN clause query and a hashtable for lookup. This reduces N queries to just 1 query.
```powershell
function Get-UserDisplayNameById {
    [CmdletBinding()]
    param(
        [Parameter(Mandatory = $true)]
        [string[]]$UserIds,
        [Parameter(Mandatory = $true)]
        $Database
    )

    # 1. Handle edge cases
    if ('All' -in $UserIds) { return 'All' }
    
    # Filter out nulls/empty strings
    $validIds = $UserIds | Where-Object { -not [string]::IsNullOrWhiteSpace($_) }
    if ($validIds.Count -eq 0) { return @() }

    # 2. Construct a single batch query
    # Format IDs for SQL IN clause: 'id1','id2','id3'
    $formattedIds = $validIds | ForEach-Object { "'$($_ -replace "'", "''")'" }
    $inClause = $formattedIds -join ','

    # We need to select 'id' as well to map the results back to the input list
    $query = "SELECT id, displayName FROM User WHERE id IN ($inClause)"
    $dbResults = Invoke-DatabaseQuery -Database $Database -Sql $query

    # 3. Create a lookup hashtable for O(1) access
    $userLookup = @{}
    if ($dbResults) {
        foreach ($row in $dbResults) {
            # Ensure we handle case sensitivity if needed, though IDs are usually consistent
            $userLookup[$row.id] = $row.displayName
        }
    }

    # 4. Map results back to original order
    $UserResult = @()
    foreach ($id in $validIds) {
        if ($userLookup.ContainsKey($id) -and -not [string]::IsNullOrWhiteSpace($userLookup[$id])) {
            $UserResult += $userLookup[$id]
        } else {
            # Fallback to ID if user not found in DB
            $UserResult += $id
        }
    }

    return $UserResult
}
```

alexandair (changes requested):
@komalp2025 Please, address my feedback.

src/powershell/tests/Test-Assessment.25371.ps1
komalp2025: This code is already using Get-ZtConditionalAccessPolicy 

image

src/powershell/tests/Test-Assessment.25371.ps1
```diff
@@ -0,0 +1,159 @@
...
+    # Prerequisite Check: If Q1 shows signalingStatus is not enabled, the check is Not Applicable
+
+    $passed = $true
+    $testResultMarkdown = ''
+    if (-not $gsaSettings -or $gsaSettings.signalingStatus -ne 'enabled') {
+        $passed = $false
+        $testResultMarkdown = "ℹ️ Global Secure Access with Conditional Access signaling is not configured in this tenant. Universal CAE is not applicable.`n`n%TestResult%"
+    }
+    else {
+        $passed = $ContinuousAccessEvaluationDisabledPolicies.Count -eq 0
```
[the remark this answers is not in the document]
komalp2025: Test is started as passed 
image

alexandair (approved):
LGTM

PR #703, opened 2025-12-12
System - Adding certificate authentication to Connect-ZtAssessment

FriedrichWeinmann:
While previously established connections using certificate-based authentication (or other application-mode authentication options) were already supported, doing so directly from within `Connect-ZtAssessment` was not.

Specifically, this update makes a few changes:

+ Added `-Certificate` parameter to `Connect-ZtAssessment`. It accepts a certificate object, a thumbprint or a subject.
+ Added graph session verification to `Connect-ZtAssessment` - after connecting, it will now validate immediately, whether the session is good enough for our toolkit (scopes, role membership).
+ Refactored `Test-ZtContext`, for cleaner code and including validation errors in the logs


merill (approved):
LGTM

PR #717, opened 2025-12-19
Network - 25383: Global and GSA admin privileges are tightly limited to prevent tenant-wide compromise  

ashwinikarke:
Added the test file and the .md file. 
Updated the implementation in line with the revised spec.


sandeepjha000 (commented):
please address the comments

alexandair (approved):
LGTM

PR #720, opened 2025-12-22
35002 - Add test for Cross-Tenant Access Policy (XTAP) RMS settings

alexandair:
Fix https://github.com/microsoft/ztspecs/issues/57


SagarSathe (approved):
LGTM

PR #721, opened 2025-12-22
35003 - Add test for sensitivity label configuration and reporting

alexandair:
Fix https://github.com/microsoft/ztspecs/issues/54


SagarSathe (approved):
LGTM

PR #732, opened 2025-12-28
35008 - Add test for SPO Default Site Label (Tenant-Wide)

alexandair:
Fix https://github.com/microsoft/ztspecs/issues/55


SagarSathe (approved):
LGTM

PR #769, opened 2026-01-12
Network-25408: updated code to use helper function

sandeepjha000:
updated code to use helper function Find-ZtProfilesLinkedToPolicy.ps1


alexandair (approved):
LGTM

SagarSathe (approved):

PR #848, opened 2026-02-02
Data-35037:  Purview Audit Logging Enabled

sandeepjha000:
adding test 35037


merill (changes requested):
@sandeepjha000 if there is an error getting the logging status, we should flag the test as skipped and provide the reason.

Right now, it doesn't get a pass or fail based on the logic.

alexandair:
Passed my review. It's expected that "skipped reason" is fixed in a bulk change in the near future.

LGTM

/cc @merill

merill (approved):

PR #862, opened 2026-02-05
Network 25539 IDPS Inspection is Enabled in Deny Mode on Azure Firewall

komalp2025:
https://github.com/microsoft/ztspecs/issues/46


alexandair (changes requested):
@komalp2025
Does the response to the initial Query 1 include the `sku` and `intrusionDetection` properties?
Recommendation: Check if `$policies` (from the first call) already contains the required data. If so, removing the inner loop and second API call will significantly speed up the test for customers with many policies.

komalp2025:
> @komalp2025 Does the response to the initial Query 1 include the `sku` and `intrusionDetection` properties? Recommendation: Check if `$policies` (from the first call) already contains the required data. If so, removing the inner loop and second API call will significantly speed up the test for customers with many policies.

Hi @alexandair , I raised this redundancy with Andrew already. According to him "From the above the second API call does not have the IDS settings. If we use only this, we might miss some Firewall policies that have IDS enabled. Please use the steps as they are." This was discussed in spec comment section.

alexandair (approved):
LGTM

PR #867, opened 2026-02-06
Get-LabelPolicySettings - update

praneeth-0000:
Updated the function to include attachmentaction property. This is needed for test 35014


alexandair (approved):
LGTM

PR #892, opened 2026-02-12
Network - 27001 - TLS inspection bypass policies are regularly reviewed to prevent security protection gaps

praneeth-0000:
https://github.com/microsoft/ztspecs/issues/319


alexandair (changes requested):

**Current Behavior:** The Invoke-ZtGraphRequest call is wrapped in a try/catch block that logs a warning. If the call fails (e.g., due to 403 Forbidden or API outage), $tlsInspectionPolicies remains $null. The subsequent check if (-not $tlsInspectionPolicies ...) then causes the test to return SkippedBecause NotApplicable.

**Problem:** This masks actual errors (like missing permissions) as "Not Applicable," which falsely implies the feature is not in use. "Not Applicable" should only be returned when the API successfully returns 0 items.

**Remediation:** Use the variable $errorMsg to capture the exception in the catch block. In the Assessment Logic, check if $errorMsg is populated. If it is, fail the test or report the error explicitly, rather than skipping it.


*Suggestion:* Check if `networkAccess/tlsInspectionPolicies?$expand=policyRules` is supported to fetch all data in one query, reducing execution time and API throttling risk.

alexandair (approved):
LGTM

PR #894, opened 2026-02-13
Add Header parameter Invoke-ZtGraphRequest

komalp2025:
Adding Header parameter to provide extra headers in Invoke-ZTGraphRequest. Earlier it was hardcoded to Consistency Level only.

PR results are updated at https://github.com/microsoft/ztspecs/issues/314


merill (changes requested):
@komalp2025 In this new logic, if a Header is passed it is not including the ConsistencyLevel in the header collection. We need to include it based on the value provided in the ConsistencyLevel parameter

komalp2025:
> @komalp2025 In this new logic, if a Header is passed it is not including the ConsistencyLevel in the header collection. We need to include it based on the value provided in the ConsistencyLevel parameter

Hi @merill , this is already taken care of.  Also if user has provided consistency level in header, it will also get reset to the value provided in consistency parameter. Please see below code snippet

```powershell
PS C:\GitRepo\zerotrustassessment> $requestHeaders = if ($Headers) { $Headers.Clone() } else { @{} }
PS C:\GitRepo\zerotrustassessment> $requestHeaders                                               

Name                           Value
----                           -----
Content-Type                   application/json

PS C:\GitRepo\zerotrustassessment> $ConsistencyLevel = 'eventual'
PS C:\GitRepo\zerotrustassessment> $requestHeaders['ConsistencyLevel'] = $ConsistencyLevel
PS C:\GitRepo\zerotrustassessment> $requestHeaders

Name                           Value
----                           -----
Content-Type                   application/json
ConsistencyLevel               eventual
```

```powershell
PS C:\GitRepo\zerotrustassessment>  $headers = @{"Content-Type" = "application/json" 
>> "ConsistencyLevel" = 'testvalue'}
PS C:\GitRepo\zerotrustassessment> $requestHeaders = if ($Headers) { $Headers.Clone() } else { @{} }       
PS C:\GitRepo\zerotrustassessment> $requestHeaders

Name                           Value
----                           -----
Content-Type                   application/json
ConsistencyLevel               testvalue

PS C:\GitRepo\zerotrustassessment> $requestHeaders['ConsistencyLevel'] = $ConsistencyLevel
PS C:\GitRepo\zerotrustassessment> $requestHeaders

Name                           Value
----                           -----
ConsistencyLevel               eventual
Content-Type                   application/json
```

> @komalp2025 In this new logic, if a Header is passed it is not including the ConsistencyLevel in the header collection. We need to include it based on the value provided in the ConsistencyLevel parameter

alexandair (approved):
@komalp2025 I've made some changes to the PR:

Pre-existing bug: `$GraphBaseUri` parameter is always ignored

```powershell
if (-not $GraphBaseUri) {
    if (-not $script:__ZtSession.GraphBaseUri) {
        $script:__ZtSession.GraphBaseUri = (Get-MgEnvironment -Name ...).GraphEndpoint
    }
}
$GraphBaseUri = $script:__ZtSession.GraphBaseUri  # ← ALWAYS overwrites!
```

Even if a caller passes `-GraphBaseUri https://custom.endpoint`, the very last line **unconditionally** overwrites it with the session value. The outer if correctly skips the initialization, but the reassignment destroys the caller's value. This has been broken since before this PR.

**Fix**: The last line should be else-guarded or inside the outer if:
```powershell
if (-not $GraphBaseUri) {
    ...
    $GraphBaseUri = $script:__ZtSession.GraphBaseUri
}
```

The cache key in `Invoke-ZtGraphRequestCache.ps1` is just `$Uri.AbsoluteUri`.
Before this PR, the only header was `ConsistencyLevel = eventual` (a constant), so this was harmless. **Now that callers can pass arbitrary headers**, two requests to the same URL with different `Prefer` or other headers will share the same cache entry and return incorrect data.

Example scenario:
```powershell
# Call 1 — with Prefer header, result gets cached
Invoke-ZtGraphRequest -RelativeUri "users" -Headers @{ Prefer = 'outlook.body-content-type="text"' }

# Call 2 — no Prefer header, silently returns Call 1's cached result  
Invoke-ZtGraphRequest -RelativeUri "users"
```
**Fix**: Incorporate serialized headers into the cache key.

merill (approved):
LGTM

PR #901, opened 2026-02-16
Add Get-FormattedError function

praneeth-0000:
Added `Get-FormattedError` function which formats error object from catch block and formats it so it's user friendly and readable.

Network Spec Example
- Without using the `Get-FormattedError` function
image

- With using the `Get-FormattedError` function
image

- With using the `Get-FormattedError` function `-IncludeDetails`
image

Data Spec Example
- Without using the `Get-FormattedError` function
image

- With using the `Get-FormattedError` function
image

### DIFF
--- a/src/powershell/private/core/Get-FormattedError.ps1
+++ b/src/powershell/private/core/Get-FormattedError.ps1
@@ -1,0 +1,130 @@
+function Get-FormattedError {
+    <#
+    .SYNOPSIS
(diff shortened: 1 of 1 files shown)


merill (commented):
@alexandair  can you review the code before I merge?

merill:
Closing this now as the error handling has been added by Gael.

PR #908, opened 2026-02-17
Fixing the error caused by data in wrong part of PSD1

gaelcolas:
The data was in the wrong part of the PSD1 which ended with Publish-Module to incorrectly move the entries.


alexandair (approved):
LGTM

PR #913, opened 2026-02-17
Different dependencies based on OS

gaelcolas:
fixing Save-Module when using PowerShellGet
Moving Required modules to Lazy loader


alexandair (approved):
LGTM

PR #915, opened 2026-02-18
Fixed macos loading issues and warning

merill:
**Root cause**

Our previous fix pre-loaded three shared DLLs into the default AssemblyLoadContext:
- Microsoft.IdentityModel.Abstractions.dll (needed -- fixes Graph's MissingMethodException)
- Azure.Identity.dll (harmful)
- Azure.Core.dll (harmful)

Az.Accounts uses its own Assembly Load Context (Microsoft.Azure.PowerShell.AuthenticationAssemblyLoadContext.dll registers AzAssemblyLoadContextInitializer.RegisterAzSharedAssemblyLoadContext()). This ALC isolates Az's DLLs from the default context. When we pre-loaded Azure.Identity.dll (Graph's v1.13.2.0) and Azure.Core.dll (Az's v1.50.0.0) into the default context, it interfered with Az's ALC type resolution, causing the "Entry point was not found" error during Connect-AzAccount.

**Fix**
Removed Azure.Identity.dll and Azure.Core.dll from the pre-load list. Now we only pre-load Microsoft.IdentityModel.Abstractions.dll — the specific low-level type-definition DLL that caused the original Graph MissingMethodException. This DLL defines the IIdentityLogger interface that all modules share in the default context, so loading the newest version (8.6.1.0 from Graph) is both necessary and safe.

Both Graph and Azure connections should now work.

### DIFF
--- a/src/powershell/Initialize-Dependencies.ps1
+++ b/src/powershell/Initialize-Dependencies.ps1
@@ -131,10 +131,11 @@ function Initialize-Dependencies {
         Write-Host -Object "`r`n" -ForegroundColor Yellow
         Write-Host -Object '⚠️ Warning: The ZeroTrustAssessment module is designed to run on Windows, in PowerShell 7.'
         Write-Host -Object 'Some pillars require modules that can only run on Windows PowerShell (Windows PowerShell 5.1) with implicit remoting.' -ForegroundColor Yellow
-        # skipping module installation.
+        Write-Host -Object 'The following Windows-only modules will not be available: AipService, Microsoft.Online.SharePoint.PowerShell' -ForegroundColor Yellow
         #endregion
     }
-    elseif (-not $SkipModuleInstallation.IsPresent)
+
+    if (-not $SkipModuleInstallation.IsPresent)
     {
         Write-Host -Object "`r`n"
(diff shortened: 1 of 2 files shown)


src/powershell/Initialize-Dependencies.ps1
```diff
@@ -131,10 +131,11 @@ function Initialize-Dependencies {
         Write-Host -Object "`r`n" -ForegroundColor Yellow
         Write-Host -Object '⚠️ Warning: The ZeroTrustAssessment module is designed to run on Windows, in PowerShell 7.'
         Write-Host -Object 'Some pillars require modules that can only run on Windows PowerShell (Windows PowerShell 5.1) with implicit remoting.' -ForegroundColor Yellow
-        # skipping module installation.
+        Write-Host -Object 'The following Windows-only modules will not be available: AipService, Microsoft.Online.SharePoint.PowerShell' -ForegroundColor Yellow
         #endregion
     }
-    elseif (-not $SkipModuleInstallation.IsPresent)
+
+    if (-not $SkipModuleInstallation.IsPresent)
```
gaelcolas: To me, that's the only change valid.
Everything else does not seem necessary.

I'm still trying to work it out...

PR #919, opened 2026-02-19
Fixing on MacOS

gaelcolas:
On WSL ubuntu there are some unresolved issues:
- EXO module seems to require being run elevated
- Elevated via sudo does not work with prompting for creds in a web page


merill (approved):
LGTM tested on macOS

PR #76, opened 2024-07-02
Update RMD_060.md

Clay-Microsoft:

PR #926, opened 2026-02-20
Data - 35028 - Email retention policies configured

sandeepjha000:
Data - 35028 - Email retention policies configured

[Issue #76](https://github.com/microsoft/ztspecs/issues/76)

[Spec-35028](https://github.com/microsoft/ztspecs/blob/main/specs/data/35028.md)

### DIFF
--- a/src/powershell/tests/Test-Assessment.35028.ps1
+++ b/src/powershell/tests/Test-Assessment.35028.ps1
@@ -48,7 +48,8 @@ function Test-Assessment-35028 {
         Write-PSFMessage "Querying retention compliance policies" -Level Verbose
         # These cmdlets are available only in Security & Compliance PowerShell
         # Reference: https://learn.microsoft.com/en-us/powershell/module/exchangepowershell/get-retentioncompliancepolicy?view=exchange-ps
-        $retentionPolicies = Get-RetentionCompliancePolicy -ErrorAction Stop
+        # Note: -DistributionDetail is required for accurate ExchangeLocation values
+        $retentionPolicies = Get-RetentionCompliancePolicy -DistributionDetail -ErrorAction Stop
 
         # Q2: Get retention rules associated with policies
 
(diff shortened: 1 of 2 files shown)


src/powershell/tests/Test-Assessment.35028.ps1
```diff
@@ -0,0 +1,173 @@
...
+    $customStatus = $null
+
+    if ($errorMsg) {
+        Write-PSFMessage "Query failure or connection issue: $errorMsg" -Level Warning
+        $testResultMarkdown = "⚠️ Unable to determine retention policy configuration due to permissions issues or query failure.`n`n%TestResult%"
+        $customStatus = 'Investigate'
+    }
+    else {
+        # Filter for enabled retention policies with Exchange workload
+        $enabledExchangePolicies = @($retentionPolicies | Where-Object { $_.Enabled -eq $true -and $_.Workload -match 'Exchange' })
```
alexandair: The [official Microsoft documentation](https://learn.microsoft.com/en-us/powershell/module/exchange/get-retentioncompliancepolicy?view=exchange-ps) explicitly states:

> **Workload**: Currently, it shows all workloads (Exchange, SharePoint, OneDriveForBusiness, Skype, ModernGroup). **It doesn't show the actual workload on which the policy is applied.**

This means `$_.Workload -match 'Exchange'` will match **every policy** regardless of its actual workload scope. A SharePoint-only policy would still match. Consequently:
- The test could **false-pass** if any enabled retention policy exists, even if none targets Exchange
- The report table's "Workload" column would show identical values for all policies

**Fix:** Use the `ExchangeLocation` property to determine actual Exchange scope. You may need `-DistributionDetail` for accurate location values:
```powershell
$retentionPolicies = Get-RetentionCompliancePolicy -DistributionDetail -ErrorAction Stop
$enabledExchangePolicies = @($retentionPolicies | Where-Object {
    $_.Enabled -eq $true -and
    $null -ne $_.ExchangeLocation -and
    @($_.ExchangeLocation).Count -gt 0
})
```

The same filtering issue affects the Report Generation region:
```powershell
$exchangePolicies = @($retentionPolicies | Where-Object { $_.Workload -match 'Exchange' })
```


###  Suggestion
#### Consider Using `-DistributionDetail` Documented Caveat

The docs state: *"If you don't use this [DistributionDetail] switch, the values of the DistributionStatus and \*Location property values will be inaccurate."*

If you switch to checking `ExchangeLocation`, you'll need `-DistributionDetail` for accurate values. Be aware this makes the query slower for large tenants.

alexandair (changes requested):

alexandair (approved):
LGTM

SagarSathe (approved):

PR #968, opened 2026-02-27
BugFIX - 25550 - flag as failed for a tenant that has no Azure Firewall

### LINKED ISSUE #961: Netspec 25550 - flag as failed for a tenant that has no Azure Firewall
My tenant does not have any AzNet capabilities enabled. We fail the Az Firewall check incorrectly, instead of skipping it:

Image

The other AzNet checks are treated correctly:

Image

Manoj-Kesana:
Fix #961

### DIFF
--- a/src/powershell/tests/Test-Assessment.25550.ps1
+++ b/src/powershell/tests/Test-Assessment.25550.ps1
@@ -112,6 +112,8 @@ resources
         }
         catch {
             Write-PSFMessage "TLS rule query failed: $($_.Exception.Message)" -Tag Test -Level Warning
+            Add-ZtTestResultDetail -SkippedBecause NotSupported
+            return
         }
     }
 
@@ -149,7 +151,7 @@ resources
 
     if ($premiumPolicies.Count -eq 0) {
         # Policies exist but none are Premium → Fail
-        $testResultMarkdown = "TLS inspection is not properly configured on Azure Firewall. Either the global certificate authority is missing, or no application rules have TLS inspection enabled.`n`n%TestResult%"
+        $testResultMarkdown = "❌ TLS inspection is not properly configured on Azure Firewall. Either the global certificate authority is missing, or no application rules have TLS inspection enabled.`n`n"
     }
     elseif ($attachedResults.Count -eq 0) {
         # All Premium policies are not attached to any firewall → Skipped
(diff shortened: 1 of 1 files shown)


src/powershell/tests/Test-Assessment.25550.ps1
```diff
@@ -180,7 +180,9 @@ function Test-Assessment-25550 {
     $testResultMarkdown = ''
 
     if ($firewallPoliciesWithTLS.Count -eq 0) {
-        $testResultMarkdown = "❌ No Azure Firewall Premium policies found in any subscription.`n`n"
+        Write-PSFMessage 'No Azure Firewall Premium policies found in any subscription.' -Tag Test -Level VeryVerbose
+        Add-ZtTestResultDetail -SkippedBecause NotApplicable
+        return
```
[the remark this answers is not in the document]
Manoj-Kesana: Already Updated the Skip logic

src/powershell/tests/Test-Assessment.25550.ps1
```diff
@@ -180,7 +180,9 @@ function Test-Assessment-25550 {
     $testResultMarkdown = ''
 
     if ($firewallPoliciesWithTLS.Count -eq 0) {
```
alexandair: Critical Issue — Skip is too broad, masks a spec-required Fail case

The `$firewallPoliciesWithTLS` array is only populated from **Premium** policies. When `.Count -eq 0`, there are **two** distinct scenarios:

| Scenario | Spec says | PR does |
|----------|-----------|---------|
| No firewall policies at all in any subscription | **Skipped** | ✅ Skipped (correct) |
| Firewall policies exist but **none are Premium** | **Fail** (spec eval step 5) | ❌ Skipped (incorrect) |

The spec explicitly states (eval step 5):
> *"If `premiumPolicies` is empty → **Fail** (firewall policies exist but no Premium SKU found)"*

**Fix:** Track whether *any* firewall policies were found across all subscriptions, then differentiate:

```powershell
# Before the subscription loop, add:
$anyFirewallPoliciesFound = $false

# Inside the subscription loop, after collecting $fwPolicies:
if ($fwPolicies.Count -gt 0) {
    $anyFirewallPoliciesFound = $true
}

# In assessment logic, replace the single if-block:
if ($firewallPoliciesWithTLS.Count -eq 0 -and -not $anyFirewallPoliciesFound) {
    Write-PSFMessage 'No Azure Firewall policies found in any subscription.' -Tag Test -Level VeryVerbose
    return
}
elseif ($firewallPoliciesWithTLS.Count -eq 0) {
    # Policies exist but none are Premium — this is a fail
    $testResultMarkdown = "❌ Azure Firewall policies were found but none use the Premium SKU required for TLS inspection.`n`n"
}
```

There is also this issue (pre-existing, not introduced by this PR):

**⚠️ Missing `properties.firewalls` check (spec eval step 5–6)**

The spec requires checking whether a Premium policy is attached to an actual firewall:
> *"Check if `properties.firewalls` exists and is non-empty. If empty or missing → skip this policy."*

The code never checks `$policyDetail.properties.firewalls`. Unattached policies are evaluated as if they're active, which could cause false passes or false fails on orphaned policies.

src/powershell/tests/Test-Assessment.25550.ps1
```diff
@@ -149,7 +151,7 @@ resources
 
     if ($premiumPolicies.Count -eq 0) {
         # Policies exist but none are Premium → Fail
-        $testResultMarkdown = "TLS inspection is not properly configured on Azure Firewall. Either the global certificate authority is missing, or no application rules have TLS inspection enabled.`n`n%TestResult%"
+        $testResultMarkdown = "❌ TLS inspection is not properly configured on Azure Firewall. Either the global certificate authority is missing, or no application rules have TLS inspection enabled.`n`n"
```
alexandair: What happened to `%TestResult%`?

PR #979, opened 2026-03-02
Network 25400: Is Port 53 published and private DNS is not used

ashwinikarke:


ashwinikarke:
Raised another PR :  https://github.com/microsoft/zerotrustassessment/pull/998

PR #985, opened 2026-03-03
Network - 27015 - HTTP DDoS Protection Ruleset is Enabled in Application Gateway WAF

praneeth-0000:
[HTTP DDoS Protection Ruleset is Enabled in Application Gateway WAF](https://github.com/microsoft/ztspecs/issues/462)


alexandair (approved):
LGTM

PR #999, opened 2026-03-05
Network - 27000 - High-risk WCF categories (Criminal activity, Hacking, Illegal software) are blocked

sandeepjha000:
[Spec-27000](https://github.com/microsoft/ztspecs/blob/main/specs/network/27000.md)
[Issue #296](https://github.com/microsoft/ztspecs/issues/296)


alexandair:
Consider extracting the per-category evaluation into a helper**

The per-category evaluation block (lines ~126-198) is ~70 lines of nested logic. Extracting it into a private function (e.g., `Get-CategoryBlockStatus`) would improve readability and testability, consistent with how `New-FailedCategoryResults` already abstracts the failure case.

PR #1000, opened 2026-03-05
Data 35041 Browser data loss prevention is enabled for AI apps via Edge for Business 

komalp2025:
https://github.com/microsoft/ztspecs/issues/132#issuecomment-3943239635


alexandair (changes requested):
@komalp2025 
We never use it like this:
Get-SafeMarkdown($policy. Name)
 
Get-SafeMarkdown is not needed everywhere. If we expect value from API like enabled/disabled or something similar, there is no need to sanitize it with Get-SafeMarkdown

alexandair (approved):
LGTM

PR #1014, opened 2026-03-06
UX Improvements

gaelcolas:
This version changes:
- Fixes the build to use the Metadata module to avoid breaking the PSD1.
- Complete refactor of Connect-ZtAssessment. Clearer messages, tracking of what's connected. Loading exact modules.
- Skip tests when the services are not connected.
- Skip tests when the CompatibleLicense is not met.
- Fine tuning the UX for Connect-ZtAssessment


merill (approved):
LGTM

PR #1015, opened 2026-03-08
fixing build with Metadata module

gaelcolas:
Fixing the build for the FileList issue.
Fixing the module version prerelease tag.


merill (approved):
LGTM

PR #1016, opened 2026-03-08
adding timeout to test run

gaelcolas:
The timeout is evaluated only when the parallel tests have been queued, but is calculated from the beginning of the test run.
The default is 24 hours but can be passed as parameter to `Invoke-ZtAssessment`.


merill (approved):
LGTM

PR #1019, opened 2026-03-09
Network - 27017 - JavaScript Challenge is Enabled in Application Gateway WAF - updated pass message

praneeth-0000:
Network - 27017 - JavaScript Challenge is Enabled in Application Gateway WAF - updated pass message


alexandair (approved):
LGTM

PR #1043, opened 2026-03-12
Network - 26880 - Request Body Inspection is enabled in Azure Front Door WAF

aahmed-spec:
Update Test-Assessment.26880.ps1 as per spec


alexandair (approved):
LGTM

PR #1044, opened 2026-03-12
Network 26879: Request Body Inspection is enabled in Application Gateway WAF - Updated Pass/Fail Logic

ashwinikarke:
This PR updates the pass/fail evaluation logic for Application Gateway WAF policies.

Previously, the check only validated whether `policySettings.requestBodyCheck` was set to `true`. The logic has now been enhanced to also verify that `policySettings.state` is `Enabled` and `policySettings.mode` is `Prevention`.

A policy will now Pass only if all three conditions are met, and Fail if any of them are not satisfied.


alexandair (approved):
LGTM

PR #1049, opened 2026-03-13
Network - 27000 - Code improvements

sandeepjha000:
- making default action of a policy to unknown, treating as 'Not blocked'
- early exit when only default policy exists(All websites)
- moved per category block status evaluation to helper function region


alexandair (approved):
LGTM

PR #1053, opened 2026-03-13
Network - 26883 - Default rule set is assigned in Azure Front Door WAF

aahmed-spec:
made changes as per updated spec


alexandair (changes requested):

alexandair (approved):
LGTM

PR #1058, opened 2026-03-14
Adding progress log + test Az is really connected

gaelcolas:
Also quick fix to call Connect when you Invoke and no service is connected.


merill (approved):
LGTM

PR #1062, opened 2026-03-16
Do not render User authentication and Device sign-ins charts when Sig…

### LINKED ISSUE #1050: Handle case when Sign-In table does not exist

sandeepjha000:
When Sign-In table does not exist:
- do not render User authentication & Device sign-ins
- Fixes #1050


merill (approved):
LGTM

PR #1070, opened 2026-03-16
Display full module version in startup banner (fixes #1065)

### LINKED ISSUE #1065: Assessment Version Number suggestion
It would be helpful if the assessment script showed the correct version details when it starts, it currently shows V2 when it is actually V2.xx. This would help us know that we about to run a new version that we are expecting to run. 

Image

alexandair:
Fix #1065


merill (approved):
LGTM

PR #1083, opened 2026-03-18
Network 25411: TLS inspection is enabled and correctly configured for outbound traffic - Spec Update

### LINKED ISSUE #962: Netspec 25411 - Incorrectly flags TLSi settings
We fail this test:
Image

even though my tenant has policies configured:

Image

ashwinikarke:
PR Contains below changes:

1. Collect all linked profiles for table display regardless of state (previously only enabled profiles were collected, causing disabled entries to silently disappear from tables)
2. Tables now show profiles with `❌ Disabled` state so users can see misconfigured entries
3. Added `TLSPolicyLinkState -eq 'enabled'` check to Baseline Profile pass/fail logic (previously a disabled policy link would incorrectly count as a pass)
4. Added `TLSPolicyLinkState -eq 'enabled'` check to Security Profile pass/fail logic for consistency
5. Renamed `$enabledBaseLineProfiles` collection variable to `$baselineProfileResults` to reflect it now holds all linked baseline profiles, not only passing ones
6. Pass/fail filtering now derived post-collection via `Where-Object` on `$baselineProfileResults` and `$securityProfileResults`

### DIFF
--- a/src/powershell/tests/Test-Assessment.25411.ps1
+++ b/src/powershell/tests/Test-Assessment.25411.ps1
@@ -131,7 +131,7 @@ function Test-Assessment-25411 {
     if ($baselineProfileResults.Count -gt 0) {
 
         $mdInfo += "`n## TLS Inspection Policies Linked to Baseline Profiles`n`n"
-        $mdInfo += "| Linked Profile Name | Linked Profile Priority | Linked Policy Name | Policy Link State | Profile State |`n"
+        $mdInfo += "| Linked profile name | Linked profile priority | Linked policy name | Policy link state | Profile state |`n"
         $mdInfo += "| :--- | :--- | :--- | :--- | :--- |`n"
         foreach ($policy in $baselineProfileResults) {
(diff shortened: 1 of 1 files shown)


src/powershell/tests/Test-Assessment.25411.ps1
alexandair: Unused `$expand=conditionalAccessPolicies` in filteringProfiles query**

The filteringProfiles request includes:
```
'$expand' = 'policies(...),conditionalAccessPolicies($select=id,displayName)'
```
However, `Find-ZtProfilesLinkedToPolicy` **never reads** the `conditionalAccessPolicies` data from the expand. CA policy matching is done entirely by querying `$allCAPolicies` and filtering by `sessionControls.globalSecureAccessFilteringProfile.profileId`. This fetches data that is silently discarded. The expand should be removed to reduce payload size and avoid confusion.

alexandair (changes requested):

alexandair (approved):
LGTM

SagarSathe (approved):

PR #1084, opened 2026-03-18
Fix: Update Q3 notification query to use Admin assignment in test 21820

praneeth-0000:
Updates the Query 3 in test 21820 to use Notification_Admin_EndUser_Assignment in place of Notification_Requestor_EndUser_Assignment and changes the fail scenario logic, as mentioned in issue #1039.

PR #1085, opened 2026-03-18
Fix: Update Q3 notification query to use Admin assignment in test 21819

praneeth-0000:
Similar to PR #1084 , 21819 also has same issue. This PR updates the Query 3 to use Notification_Admin_EndUser_Assignment in place of Notification_Requestor_EndUser_Assignment.


alexandair (approved):
LGTM

SagarSathe (approved):

SagarSathe (approved):

PR #1089, opened 2026-03-19
Sorting order Fix

### LINKED ISSUE #991: Sort order should be High, Medium, Low
When the report opens after the run and I navigate to the Network Pillar, the results are not ordered by Status. See below. I'd recommend we have a default filter by Status.

Image

Manoj-Kesana:


src/report/src/components/test-table/data-table.tsx
```diff
@@ -79,8 +74,8 @@ export function DataTable<TData extends Test, TValue>({
         TestSfiPillar: false,
         // Hide TestMinimumLicense by default
         TestMinimumLicense: false,
-        // Category visible for Devices, hidden for Identity
-        TestCategory: pillar === "Devices" ? true : false,
+        // Hide TestCategory by default
+        TestCategory: false,
```
merill: Why was this changed? Category should be shown for Devices pillar by default.

merill (changes requested):
Please see recommendations.

merill (changes requested):
@Manoj-Kesana In both Identity and Network pillar I can see Skipped items between the Passed ones (see screenshot below).

On a seperate note, I don't think we need to sort in both PowerShell and in the report. We should only do the sorting in the report.

Image

merill (approved):
LGTM

PR #1111, opened 2026-03-27
Identity - 21824 - fixed case when no filteredCApolicies exist ( Bugfix/issue 779)

### LINKED ISSUE #779: 21824 - Check will pass even when there are no qualifying CA Policies
Noticed while testing ZTA in my sandbox

**Steps to reproduce:**
- Run the test in an environment with no Conditional Access policies targeting includeGuestsOrExternalUsers
 
**Expected behavior:**
The test should fail when no policies enforce sign-in frequency for guests.

**Actual behavior:**
The test passes incorrectly.

**Root cause:**

1. When no policies target `includeGuestsOrExternalUsers`, `$filteredCAPolicies` is null/empty
2. `$matchedPolicies` is derived from `$filteredCAPolicies`, so it's also null/empty
3. The final validation checks `$filteredCAPolicies.Count -eq $matchedPolicies.Count`
4. Since `0 -eq 0` (or `$null -eq $null`), the condition evaluates to `$true` and the test passes

**Suggested Fix:**
Validate if at least one policy exists

https://github.com/microsoft/zerotrustassessment/blob/6aef9c434fe59d35e93962c9a5e8c2b4b46d1c06/src/powershell/tests/Test-Assessment.21824.ps1#L56

sandeepjha000:
Fixes #779

### DIFF
--- a/src/powershell/tests/Test-Assessment.21824.ps1
+++ b/src/powershell/tests/Test-Assessment.21824.ps1
@@ -53,7 +53,7 @@ function Test-Assessment-21824 {
     $passed = $false
     $testResultMarkdown = ""
 
-    if ($filteredCAPolicies.Count -eq $matchedPolicies.Count) {
+    if ($filteredCAPolicies.Count -gt 0 -and $filteredCAPolicies.Count -eq $matchedPolicies.Count) {
         $passed = $true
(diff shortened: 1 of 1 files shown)


merill (approved):
lgtm

Issue #1100, opened 2026-03-24
Skipped Test - License should be checked before Service connection


merill:
Currently the skipped test show the reason as 'Not connected to service' even though the test is not applicable to the tenant due to it's licensing status.

This gives the false impression that the test is applicable to the users tenant.

Expected:
- Test skipping should first check if a test is valid for a tenant (based on the tenant's license)
- Only if the license is valid should the test be validated for Service connection

gaelcolas:
This has been implemented in Dev with #1115

PR #1115, opened 2026-03-27
implementing error status

gaelcolas:
- Implemented Error Management.
- fixing #1100 with first checking for license before checking for service connection.

Now the output during errors looks like this:
image

And the errored entries have the status error (as opposed to failed like failed tests).
image

The error contains an indication of the failed test and the Exception message.
Below we show the stack trace (the usual output from PowerShell's `$_ | Get-Error`)
image

Now timeouts also show as `error` instead of `failed`.
Logs should be clearer as well.
This also fixes the "disappearing" tests when they returned an error (they were not counted in total when they errored out).


merill (approved):
LGTM

Issue #1097, opened 2026-03-23
ZeroTrustAssessment incorrectly blocks execution under WDAC due to session LanguageMode validation instead of trust-based validation


vimalv4u:
Hi team,

I’ve run into an issue while trying to use ZeroTrustAssessment in an environment where WDAC is enabled, and scripts are properly code-signed and allowed.

The module is currently checking:

$ExecutionContext.SessionState.LanguageMode

and blocking execution if it returns ConstrainedLanguage.

The challenge is that under WDAC, PowerShell sessions will often report ConstrainedLanguage even when:

The script is signed
The signing certificate is trusted
The WDAC policy explicitly allows execution

In this setup, trusted scripts are still able to run with full capability, but the module exits early based purely on the session LanguageMode.

Is there a reason the module depends only on the session LanguageMode instead of checking if the script is trusted or permitted under WDAC? Asking customers to identify a server that doesn't meet the compliance level to run this module appears to be a catch-22 situation.

Is there a supported way to run this module in a WDAC-controlled environment where scripts are signed and permitted, but the session still reports ConstrainedLanguage?

PR #1127, opened 2026-04-02
fix: add -IgnoreLanguageMode switch to bypass CLM check in WDAC-managed environments

praneeth-0000:
addresses issue #1097 

## Summary
In environments where Windows Defender Application Control (WDAC) or AppLocker is enabled, PowerShell sessions always report `ConstrainedLanguage` as the session language mode, even when the module is explicitly trusted by policy and runs with full .NET capability. The module was performing a hard block on any session reporting non-Full language mode, making it impossible to run the assessment on WDAC-managed devices regardless of trust configuration.

## Root Cause
`Test-ZtLanguageMode` reads `$ExecutionContext.SessionState.LanguageMode` and immediately exits with a terminating error if the value is not `FullLanguage`. The session-level `LanguageMode` property is set at process startup by WDAC and never changes.

## Fix
Added an `-IgnoreLanguageMode` switch to `Connect-ZtAssessment` and `Invoke-ZtAssessment`, passed through to `Test-ZtLanguageMode`. When specified:

- The CLM detection produces a yellow warning instead of a terminating red error
- The function returns $true and the assessment proceeds normally
- All existing behavior is unchanged when the switch is not used

## Usage

**Before (blocks in WDAC environments):**
```
Connect-ZtAssessment
Invoke-ZtAssessment
```

**After (use in WDAC-managed environments where the module is trusted):**
```
Connect-ZtAssessment -IgnoreLanguageMode
Invoke-ZtAssessment
```

## Notes

- This switch is intentionally opt-in. Without it, behavior is unchanged and CLM still blocks by default.
- If the module is not trusted by the WDAC policy and CLM is truly enforced, the assessment may fail mid-run with confusing errors. The warning message communicates this risk to the user.

## Screenshots
`Connect-ZtAssessment` with and without `-IgnoreLanguageMode` switch
image

---
Update
Added code to 
- Remember user's choice to Ignore Language Mode when user does Connect-ZtAssessment
- Show warning message only once

image


src/powershell/public/Invoke-ZtAssessment.ps1
```diff
@@ -237,7 +249,7 @@ $titleLine
 	#region Preparation
 	Show-ZtiBanner
 
-	if (-not (Test-ZtLanguageMode)) {
+	if (-not (Test-ZtLanguageMode -IgnoreLanguageMode:$IgnoreLanguageMode)) {
 		Stop-PSFFunction -Message "PowerShell is running in Constrained Language Mode, which is not supported." -EnableException $true -Cmdlet $PSCmdlet
```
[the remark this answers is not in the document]
praneeth-0000: fixed

src/powershell/public/Invoke-ZtAssessment.ps1
```diff
@@ -57,6 +57,12 @@ Tests that exceed this limit are recorded as timed out and execution continues w
 For Data pillar tests and other external-module/remoting-heavy operations, timeout is a
 best-effort interruption rather than a guaranteed hard stop of the underlying operation.
 
+.PARAMETER IgnoreLanguageMode
```
alexandair, line 60: If a user runs commands separately:

    Connect-ZtAssessment -IgnoreLanguageMode   # succeeds, CLM warning shown
    Invoke-ZtAssessment                         # BLOCKS — CLM hard-stop

`Invoke-ZtAssessment` skips auto-connect (services already connected) but still calls `Test-ZtLanguageMode` **without** `-IgnoreLanguageMode`. The user acknowledged CLM risk during connect but must redundantly pass the switch again.

**Suggestion:** Persist the decision in a script-scope variable:

    # In Connect-ZtAssessment, after successful Test-ZtLanguageMode:
    if ($IgnoreLanguageMode) { $script:IgnoreLanguageMode = $true }

    # In Invoke-ZtAssessment, before Test-ZtLanguageMode:
    $effectiveIgnore = $IgnoreLanguageMode -or $script:IgnoreLanguageMode
    if (-not (Test-ZtLanguageMode -IgnoreLanguageMode:$effectiveIgnore)) { ... }

src/powershell/public/Invoke-ZtAssessment.ps1
```diff
@@ -181,11 +187,17 @@ function Invoke-ZtAssessment {
...
+		# When specified, bypasses the Constrained Language Mode check. Use only in WDAC-managed environments
+		# where the session reports CLM but the module is trusted and runs with full capability.
+		[Parameter(ParameterSetName = 'Default')]
+		[switch]
+		$IgnoreLanguageMode
 	)
 
 	if ($script:ConnectedService.Count -le 0) {
-		Connect-ZtAssessment
+		Connect-ZtAssessment -IgnoreLanguageMode:$IgnoreLanguageMode
```
alexandair, line 200: `Invoke-ZtAssessment` calls `Test-ZtLanguageMode` twice:

1. **Indirectly** via `Connect-ZtAssessment`
2. **Directly** in `#region Preparation` (line ~252)

When `Connect-ZtAssessment` succeeds with `-IgnoreLanguageMode`, the second call shows the same yellow warning **again**. This is noisy UX. Consider removing the redundant second call or suppressing the warning on repeat.

src/powershell/public/Invoke-ZtAssessment.ps1
alexandair, line 60: Not added to ConfigurationFile configurable parameters

`Invoke-ZtAssessment` supports loading parameters from JSON config:

    $configurableParameters = @('Path', 'Days', 'MaximumSignInLogQueryTime',
        'ShowLog', 'ExportLog', 'DisableTelemetry', 'Resume', 'Tests', 'TestTimeout')

`IgnoreLanguageMode` is not in this list. For WDAC-managed environments running unattended assessments via config files, the switch cannot be set declaratively. Consider adding it.

alexandair (changes requested):

src/powershell/public/Invoke-ZtAssessment.ps1
praneeth-0000: Thanks for the suggestion Alex, I've used scoped variables to store user's choice and Invoke-ZtAssessment doesn't need -IgnoreLanguageMode switch again.

src/powershell/public/Invoke-ZtAssessment.ps1
praneeth-0000: I've used another session variable to check if the warning is already shown and the fix now only shows warning once.

src/powershell/public/Invoke-ZtAssessment.ps1
praneeth-0000: As discussed, adding IgnoreLanguageMode to configurable parameters would not work because the json file loads after CLM kicks in, So module exists out before parsing config file.

alexandair (approved):
LGTM

PR #1135, opened 2026-04-08
Add DuckDB Cross Platform Support

billmcilhargey:
Add architecture-specific DuckDB library handling and improve error messages

Add DuckDB support for Linux x64 and Linux arm64 platforms

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -407,3 +407,6 @@ ZeroTrustReport
 /src/powershell/export
 ZeroTrustReport
 /code-tests/TestResults
+
+# Runtime-generated: copied from arch-specific subdir by Initialize-Dependencies.ps1
+/src/powershell/lib/libduckdb.so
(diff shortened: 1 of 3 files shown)


billmcilhargey:
@merill ready for review - fyi GitHub Copilot failed to review most likely due to .so's provided

if required i can remove those and then as part of the review process they would need to be injected from the team from the source

https://github.com/duckdb/duckdb/releases

For platforms Linux x64 and Linux arm64 for the same version used (wanted) as the Windows side and then dropped into the folders created since they are named the exact same.  I use a copy method in future, TODO, symlink, to detect the Linux platform used, to select the binary for DuckDB to be used and copy the proper binary a folder higher. Also as another TODO, would be to re0detect the platform at run to ensure the proper binary is being used in case the entire folder was copied to another box and the platform for Linux was changed from Linux x64 to arm64 or vice-versa and that binary in higher folder is still referenced and it would cause an issue. The workaround is to just delete it and the script will fix itself today. (.gitignroe does not sync this binary).

Thank you again!

billmcilhargey:
shifting to dev branch

PR #1140, opened 2026-04-09
Deprecating tests 25378 and 35002

### LINKED ISSUE #974: 25378 in conflict with 35002?
There are several checks that look at cross-tenant settings, and it seems some of them are in conflict with each other.

For instance it appears that _25378 - External collaboration is governed by explicit Cross-Tenant Access Policies_ works against _35002 - Cross-tenant access settings configured to allow encrypted content sharing)_

They are also related to _21790 - Outbound cross-tenant access settings are configured_ so that might also be a conflict.

sandeepjha000:
Tests 25378, 35002 and 21790 conflicted each other. Hence, we are considering only 21790 to cover the security posture and deprecating the other two tests.

Fixes #974


alexandair (approved):
LGTM

PR #1157, opened 2026-04-16
Fix #987 - Hide skipped tests by default and tests count consistent

### LINKED ISSUE #987: ZT assessment overview page number if checks is not consistent with pillar pages
After running the assessment, the overview pages shows this: 

Image

So network pillar says 23/38 tests, but then in the network pillar sub page it says. 

Image

So 54 tests in total. Shouldn´t the numbers be the same?

Manoj-Kesana:
### Problem
- Skipped tests were always visible in the test table, adding noise to the assessment results.

### Changes
- **Skipped tests are now hidden by default** - the test table filters out tests with `"Skipped"` status on initial load, giving a cleaner view of actionable results.
- **Added a "Show skipped tests" checkbox** to let users optionally reveal skipped tests when needed.
- Positioned the checkbox directly below the "Showing X of Y tests" counter, right-aligned, so it doesn't clash with the SFI pillar filter pills.
- Added a `hasSkippedTests` memo that checks if any tests in the current pillar have `"Skipped"` status.
- The checkbox only renders when skipped tests actually exist in the pillar - if there are none, it stays hidden.


astaykov:
hey @Manoj-Kesana, 
When I ran the assessment on my tenant, with the these changes, I got an error in the generated HTML report:
image

Attached is the whole result folder. Can you inspect what went wrong? The issue is most probably not be related to this change, but it is still an issue and can't validate the change, because the report is not rendering.
[dayzure.zip](https://github.com/user-attachments/files/27514529/dayzure.zip)

merill (approved):
LGTM

PR #1163, opened 2026-04-17
fix #1122 - UI progress page duplicate 

### LINKED ISSUE #1122: UI progress page duplicate fix
when we re-run the assessment the 2nd time the 1st run progress page is also getting enabled. need to make unique for each progress

Manoj-Kesana:
### Problem

**Progress dashboard:** When running the assessment multiple times in a session, the progress page from run 1 would start displaying run 2's data. Reloading the old tab also re-activated it, causing duplicate live dashboards tracking the same run.

#### Progress dashboard - unique run isolation

**Files:** `Start-ZtProgressServer.ps1`, `progress.html`

- Generate a unique `runId` (GUID) each time the progress server starts
- Store the `runId` in the shared `ProgressState` dictionary and include it in the `/api/progress` JSON response
- Pass the `runId` to the browser via URL hash fragment (`http://localhost:8924#run=<guid>`) instead of injecting into HTML
- The browser reads `runId` from `location.hash`, which **survives page reloads**
- On each poll, the dashboard compares its `runId` against the server's - if they differ, it stops polling and shows "A new assessment run has started. This tab is no longer active."

### DIFF
--- a/src/powershell/assets/progress.html
+++ b/src/powershell/assets/progress.html
@@ -335,8 +335,9 @@
 <script>
 (function() {
   const POLL_INTERVAL = 500;
-  // Read runId from URL hash (e.g. #run=abc123) so it survives page reloads
-  const MY_RUN_ID = (location.hash.match(/run=([a-f0-9]+)/i) || [])[1] || '';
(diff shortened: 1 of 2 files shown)


merill (approved):
LGTM

PR #1165, opened 2026-04-17
Feature: Add app-only CBA support for all services

### LINKED ISSUE #684: Request New Feature : Allow custom app for Entra Authentication
Dear Team,
I would like to submit a feature request regarding the Zero Trust Assessment process. Currently, the assessment relies on the Microsoft Graph Command Line for execution. While this approach works, it limits flexibility for organizations that prefer to integrate the assessment into their own workflows or automation pipelines.
Requested Feature:
Enable the possibility to use a custom application to execute the Zero Trust Assessment instead of being restricted to the Microsoft Graph Command Line. This would allow organizations to:

Integrate the assessment into existing security tools and dashboards.
Automate reporting and remediation steps using internal scripts or applications.
Maintain compliance with internal security and operational standards.

This enhancement would significantly improve adoption and operational efficiency for enterprises implementing Zero Trust principles.
Thank you for considering this request. Please let me know if additional details or use cases are needed.

### LINKED ISSUE #1148: Auto-skip tests that don't support app-only authentication
## Summary

When connecting with a custom Entra app using certificate or client secret (app-only / client credentials flow), 
some Graph APIs return `AccessDenied: Unsupported app-only call`. This causes those tests to appear as 
**failures** in the report instead of being gracefully skipped.

## Current Behavior

Tests that call delegated-only Graph APIs (e.g., `Policies/MobileDeviceManagementPolicies`) throw an 
unhandled `401 Unauthorized` error when connected app-only, causing them to show as broken/errored 
in the report.
```
WARNING: [13:54:47][Invoke-ZtRetry] Non-retryable error encountered: Response status code does not indicate success: Unauthorized (Unauthorized).. Failing immediately.
Invoke-ZtRetry: D:\Microsoft Projects\Microsoft\zerotrustassessment\src\powershell\private\core\Invoke-ZtGraphRequestCache.ps1:118
Line |
 118 |  …  $results = Invoke-ZtRetry -ScriptBlock { Invoke-MgGraphRequest -Meth …
     |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | GET https://graph.microsoft.com/beta/Policies/MobileDeviceManagementPolicies?$expand=includedGroups HTTP/2.0 401 Unauthorized Cache-Control: no-store, no-cache Vary: Accept-Encoding
     | Strict-Transport-Security: max-age=31536000 request-id: 042f205b-c01c-4a35-bb42-6e56a6655a81 client-request-id: 2bd18cc4-e8d3-44fe-877d-23de76b3f7ca x-ms-ags-diagnostic:
     | {"ServerInfo":{"DataCenter":"Central India","Slice":"E","Ring":"3","ScaleUnit":"002","RoleInstance":"PN2PEPF00000274"}} Date: Fri, 10 Apr 2026 08:24:48 GMT Content-Type: application/json
     | {"error":{"code":"AccessDenied","message":"Unsupported app-only
     | call.","innerError":{"date":"2026-04-10T08:24:49","request-id":"042f205b-c01c-4a35-bb42-6e56a6655a81","client-request-id":"2bd18cc4-e8d3-44fe-877d-23de76b3f7ca"}}}
```

## Expected Behavior

Tests that cannot run due to authentication or connection limitations should be automatically 
skipped with a clear reason rather than failing with an error. The report should still be 
generated successfully with all supported tests completing normally.

### LINKED ISSUE #1166: Issue connecting with an app registration
Hi everyone,

I am trying to run the ZTA. I created an app registration and added the needed permissions to an enterprise logic app.

Now I am trying to execute this command with the updated clientID, tenantID and Certificate (that is already upload in the app registration)
.EXAMPLE
		PS C:\> Connect-ZtAssessment -ClientID $clientID -TenantID $tenantID -Certificate 'CN=ZeroTrustAssessment' -Service Graph,Azure

		Connects to Microsoft Graph and Azure using the specified client/application ID & tenant ID, using the latest, valid certificate available with the subject 'CN=ZeroTrustAssessment'.
		This assumes the correct scopes and permissions are assigned to the application used.
	#>

but the ouput is the following without providing any information more information:

Image

Could you help me to find the issue?

Thanks,
Xavier

ashwinikarke:
Extends `-Certificate` parameter support from Graph-only to all services (Azure, Exchange Online, Security & Compliance, SharePoint Online).

Key changes:

- CBA branches added for Azure, EXO, S&C, and SPO
- SPO uses temp PFX export to work around WinPS5 cert store limitation
- AipService skips gracefully with a warning (CBA unsupported)
- Graph always prepended to service list to ensure domain inference works
- Initialize-Dependencies.ps1 pre-loads highest Microsoft.IdentityModel.Abstractions.dll to prevent DLL conflict with EXO
- Add-ZtDeviceWindowsEnrollment.ps1 skips app-only-incompatible Graph call
- Removed incorrect Remove-ZtConnectedService 'Graph'/'Azure' calls that were previously triggered by SPO admin URL inference failure

--------------------------------------------------------------------
**Bug Fix: CBA Reconnect Failures for EXO, S&C, and SPO in Same Session**
Fixes failures when reconnecting to Exchange Online, Security & Compliance, or SharePoint Online after a prior Disconnect-ZtAssessment in the same PowerShell session.

Root causes:

- `$resolvedInitialDomain` was function-local, the tenant's initial domain was resolved after connecting Graph but stored in a local variable that was lost when `Connect-ZtAssessment` returned. Any subsequent call started with $null, causing EXO/S&C CBA org lookups and SPO admin URL inference to fail even when Graph was still connected.
- Graph "already connected" branch skipped domain resolution the elseif ($isGraphConnected) branch hit continue before the domain resolution block, so EXO, S&C, and SPO had no domain when Graph was reused.
- `Disconnect-ZtAssessment` never updated `$script:ConnectedService` - after disconnecting, the internal list still contained 'Graph', causing the Graph-prepend guard to skip prepending it on the next call, so SPO ran alone with no Graph context.
- `SecurityCompliance` was never cleared on disconnect - `$serviceMap` uses `'ExchangeOnline'` as the key for the shared Disconnect-ExchangeOnline call, but SecurityCompliance is tracked separately in `$script:ConnectedService`, so it was never removed on disconnect.

Files changed:

- `Connect-ZtAssessment.ps1` - promoted $resolvedInitialDomain to $script: scope; added domain resolution in the Graph "already connected" branch; aligned EXO/S&C domain-missing error output style with SPO (direct messages + continue instead of throw)
- `Disconnect-ZtAssessment.ps1` - call Remove-ZtConnectedService after each service disconnect so the internal tracking always reflects actual state
- `Remove-ZtConnectedService.ps1` - clear $script:resolvedInitialDomain when Graph is removed so it is re-resolved on the next connect
- `Disconnect-ZtAssessment.ps1` - explicitly calls `Remove-ZtConnectedService -Service 'SecurityCompliance'` when `ExchangeOnline` is disconnected, since both are tracked separately in `$script:ConnectedService` but share a single `Disconnect-ExchangeOnline` command


docs/Connect-ZtAssessment.md
```diff
@@ -0,0 +1,129 @@
...
+| Microsoft Graph | `NetworkAccess.Read.All` |
+| Microsoft Graph | `Policy.Read.All` |
+| Microsoft Graph | `Policy.Read.ConditionalAccess` |
+| Microsoft Graph | `Policy.Read.PermissionGrant` |
+| Microsoft Graph | `PrivilegedAccess.Read.AzureAD` |
+| Microsoft Graph | `Reports.Read.All` |
+| Microsoft Graph | `RoleManagement.Read.All` |
+| Microsoft Graph | `UserAuthenticationMethod.Read.All` |
+| Office 365 Exchange Online | `Exchange.ManageAsApp` |
+| SharePoint | `Sites.FullControl.All` |
```
alexandair: Is `Sites.FullControl.All` the right permission?
It doesn't look like it's following the principle of least privilege.

src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -241,7 +248,7 @@ function Connect-ZtAssessment {
 				}
 
 				if ($Certificate) {
```
alexandair, line 267: **No parameter validation: `-Certificate` without `-ClientId`/`-TenantId`**

The documentation states all three (`-ClientId`, `-TenantId`, `-Certificate`) are required for app-only mode, but the code never validates this. Supplying `-Certificate` alone will:
- Pass `$Certificate.Certificate` to `Connect-MgGraph` without `ClientId`/`TenantId` → opaque runtime error
- Skip EXO/S&C CBA (because `$ClientId -and $Certificate` is `$false`)
- Fall through to interactive SPO connection

**Recommendation:** Add early validation at the top of the function:
```powershell
if ($Certificate -and (-not $ClientId -or -not $TenantId)) {
    throw "When -Certificate is specified, both -ClientId and -TenantId are required for app-only authentication."
}
```

src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -465,7 +487,29 @@ function Connect-ZtAssessment {
 				}
 
 				Write-Verbose -Message 'Connecting to Microsoft Exchange Online'
-				if ($UseDeviceCode) {
+				if ($ClientId -and $Certificate) {
+					# App-only CBA: requires -AppId, -Certificate, and -Organization (primary *.onmicrosoft.com domain, not a GUID).
+					$exoOrganization = $null
+					try {
+						$exoOrg = Invoke-ZtGraphRequest -RelativeUri 'organization'
```
alexandair: **Duplicate Graph API calls to `/organization` — up to 3× redundant**

The `organization` endpoint is called separately for EXO, S&C, and SPO to resolve the initial domain. If all three services are requested, this makes 3 identical Graph API calls.

**Recommendation:** Resolve the initial domain once after Graph connects successfully and cache it in a variable (e.g., `$script:InitialDomain` or local `$resolvedInitialDomain`). Reuse for EXO, S&C, and SPO.

src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -669,11 +747,6 @@ function Connect-ZtAssessment {
 					}
 				}
 			}
-			elseif(-not $adminUrl) {
```
alexandair, line 672: **Removed `elseif(-not $adminUrl)` branches silently changes behavior**

The dev branch had:
```powershell
elseif(-not $adminUrl) {
    Remove-ZtConnectedService -Service 'Graph'   # If Graph context doesn't exist
}
# ...
elseif (-not $adminUrl) {
    Remove-ZtConnectedService -Service 'Azure'   # If Azure context doesn't exist
}
```

These `Remove-ZtConnectedService` calls were arguably buggy (removing Graph/Azure status due to SPO inference failure), and the PR's change to `if` chains is logically correct — it now tries both inference methods regardless. **However**, the old code's intent (marking services as disconnected when their contexts don't exist) is now completely removed without explanation. This is likely the *correct* fix, but the PR description doesn't mention this behavioral change, making it hard to verify intent.

src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -627,10 +702,12 @@ function Connect-ZtAssessment {
 			Write-Host -Object "`nConnecting to SharePoint Online" -ForegroundColor Cyan
 			try {
 				Write-PSFMessage -Message ('Loading SharePoint Online required modules: {0}' -f ($resolvedRequiredModules.SharePointOnline.Name -join ', ')) -Level Verbose
+				$loadedSharePointOnlineModules = $null
```
alexandair: The PR adds `$loadedSharePointOnlineModules = $null` before the assignment. In dev branch, this variable was assigned directly. The explicit `$null` initialization is harmless but breaks the pattern used by every other service (Graph, Azure, EXO, AipService, S&C — none of which pre-initialize their `$loaded*Modules` variable). Either add it everywhere for consistency or remove it.

src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -465,7 +487,29 @@ function Connect-ZtAssessment {
...
+					# App-only CBA: requires -AppId, -Certificate, and -Organization (primary *.onmicrosoft.com domain, not a GUID).
+					$exoOrganization = $null
+					try {
+						$exoOrg = Invoke-ZtGraphRequest -RelativeUri 'organization'
+						$exoOrganization = $exoOrg.verifiedDomains | Where-Object { $_.isInitial } | Select-Object -ExpandProperty name -First 1
+					}
+					catch {
+						Write-Verbose -Message "Failed to resolve initial domain for Exchange Online CBA: $($_.Exception.Message)"
+					}
+					if (-not $exoOrganization) {
```
alexandair, line 532: Invoke-ZtGraphRequest -RelativeUri 'organization' error handling inconsistency
For EXO, a failure to resolve the initial domain is a hard error.

For SPO, a failure is silent — it falls through to the Azure context check. This asymmetry is confusing. If Graph is connected but the organization call fails, SPO will fail with "URL not provided" instead of a clear message about domain resolution failure.

src/powershell/private/tenantinfo/devices/Add-ZtDeviceWindowsEnrollment.ps1
```diff
@@ -14,6 +14,13 @@ function Add-ZtDeviceWindowsEnrollment
     $activity = "Getting Windows enrollment summary"
     Write-ZtProgress -Activity $activity -Status "Processing"
 
+    $mgContext = Get-MgContext -ErrorAction Ignore
+    if (-not $mgContext -or $mgContext.AuthType -eq 'AppOnly') {
```
alexandair: **`Add-ZtDeviceWindowsEnrollment` — `Get-MgContext` guard is correct but incomplete**

The guard correctly skips the incompatible API call in app-only mode. Previous Copilot review flagged the null-check, which was addressed (`-ErrorAction Ignore` + `-not $mgContext -or`). However:

- The `-not $mgContext` branch returns early even if Graph isn't connected at all (not just app-only). This means if someone somehow calls this function without connecting to Graph, it silently returns instead of failing. This might be acceptable as defensive coding, but it changes observable behavior vs. the previous code which would have attempted the call and gotten a Graph error.

src/powershell/Initialize-Dependencies.ps1
```diff
@@ -243,6 +243,54 @@ function Initialize-Dependencies {
...
+
+            $identityModelFiles = foreach ($searchDir in ($identityModelSearchDirs | Where-Object { $_ } | Sort-Object -Unique)) {
+                if (Test-Path -Path $searchDir -PathType Container) {
+                    Get-ChildItem -Path $searchDir `
+                        -Filter 'Microsoft.IdentityModel.Abstractions.dll' `
+                        -File -Force -ErrorAction SilentlyContinue
+                }
+            }
+            $bestIdentityModel = $identityModelFiles | Where-Object { $_ } |
+                Sort-Object { try { [version]($_.VersionInfo.FileVersion) } catch { [version]'0.0.0.0' } } -Descending |
```
alexandair: `FileVersion` is a string that can contain non-version text (e.g., `"8.1.0.0-preview"`; prerelease tags). The `[version]` cast will throw on these, falling back to `0.0.0.0`. Consider using `$_.VersionInfo.ProductVersion` or `[System.Diagnostics.FileVersionInfo]::GetVersionInfo($_.FullName).FileVersion` with regex extraction.

alexandair (changes requested):

src/powershell/public/Connect-ZtAssessment.ps1
ashwinikarke: Fixed. Added an early Stop-PSFFunction guard at function entry.

src/powershell/public/Connect-ZtAssessment.ps1
ashwinikarke: Fixed. The organization call is now made once immediately after Graph connects and cached in $resolvedInitialDomain. EXO, S&C, and SPO all reuse that value.

src/powershell/public/Connect-ZtAssessment.ps1
ashwinikarke: The Remove-ZtConnectedService 'Graph'/'Azure' calls were a bug. I’ve fixed it and updated the PR description to explicitly call this out.

src/powershell/Initialize-Dependencies.ps1
ashwinikarke: Fixed. Replaced $_.VersionInfo.FileVersion sort key with a regex.

src/powershell/private/tenantinfo/devices/Add-ZtDeviceWindowsEnrollment.ps1
ashwinikarke: Fixed. Split into two separate guards, each with its own accurate message: one for null context, one for AppOnly auth.

src/powershell/public/Connect-ZtAssessment.ps1
ashwinikarke: Fixed.

src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -647,63 +754,45 @@ function Connect-ZtAssessment {
...
-					Write-Verbose -Message "Failed to infer SharePoint Admin URL from Azure context: $($_.Exception.Message)"
-				}
-			}
-			elseif (-not $adminUrl) {
-				Write-Verbose -Message "No Azure context available to infer SharePoint Admin URL."
-				Remove-ZtConnectedService -Service 'Azure'
+			# Try to infer from the already-resolved initial domain (cached after Graph connected)
+			if (-not $adminUrl -and $script:resolvedInitialDomain) {
+				$tenantName = $script:resolvedInitialDomain.Split('.')[0]
+				$adminUrl = "https://$tenantName-admin.sharepoint.com"
```
alexandair: SPO Admin URL Inferred as Commercial-only for All Clouds

```powershell
if (-not $adminUrl -and $script:resolvedInitialDomain) {
    $tenantName = $script:resolvedInitialDomain.Split('.')[0]
    $adminUrl = "https://$tenantName-admin.sharepoint.com"   # ← always Commercial
}
```

The function has full `$Environment` support with a `ValidateSet` for `USGov`, `USGovDoD`, `China`. SPO admin URLs differ per cloud:
- Commercial: `https://<tenant>-admin.sharepoint.com`
- GCC High / DoD: `https://<tenant>-admin.sharepoint.us`
- China: `https://<tenant>-admin.sharepoint.cn`

This was flagged as a suppressed Copilot comment but never addressed. Users with `-Environment USGov` and no `$SharePointAdminUrl` will always get an inferred Commercial URL and a failed connection, with the error pointing them to use `-SharePointAdminUrl` rather than explaining the real cause.

**Fix:** Map the suffix from the existing `$Environment` / `$azEnvironment` value:

```powershell
$spoSuffix = switch ($Environment) {
    'USGov'    { 'sharepoint.us' }
    'USGovDoD' { 'sharepoint.us' }
    'China'    { 'sharepoint.cn' }
    default    { 'sharepoint.com' }
}
$adminUrl = "https://$tenantName-admin.$spoSuffix"
```

src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -647,63 +754,45 @@ function Connect-ZtAssessment {
...
+					if ($ClientId -and $TenantId -and $Certificate) {
+						Write-PSFMessage -Message "Using app-only CBA for SharePoint Online with ClientId '{0}'." -Level Verbose -StringValues @($ClientId)
+						# Connect-SPOService is loaded via WinPS5 implicit remoting. The -CertificateThumbprint parameter
+						# searches Cert:\CurrentUser\My inside the WinPS5 process which may not see the current session's
+						# cert store. Export the cert to a temp PFX and pass its path + password to bypass store lookup.
+						$tempPfxPath = [System.IO.Path]::Combine([System.IO.Path]::GetTempPath(), [System.IO.Path]::GetRandomFileName() + '.pfx')
+						$tempPassword = [System.Convert]::ToBase64String([System.Security.Cryptography.RandomNumberGenerator]::GetBytes(32))
+						$tempSecurePassword = ConvertTo-SecureString $tempPassword -AsPlainText -Force
+						try {
+							$pfxBytes = $Certificate.Certificate.Export([System.Security.Cryptography.X509Certificates.X509ContentType]::Pfx, $tempPassword)
```
alexandair, line 791: Security: Temp PFX file exposes private key material on disk — undocumented requirement

```powershell
$pfxBytes = $Certificate.Certificate.Export([System.Security.Cryptography.X509Certificates.X509ContentType]::Pfx, $tempPassword)
[System.IO.File]::WriteAllBytes($tempPfxPath, $pfxBytes)
Connect-SPOService ... -CertificatePath $tempPfxPath ...
```

Two issues:

1. **Exportable private key requirement is undocumented.** If the certificate was installed with "Do not allow the private key to be exported", `Certificate.Export()` throws `CryptographicException: Key not valid for use in specified state`. This error lands in the generic SPO catch block and is shown as "Failed to connect. Tests requiring SharePoint Online will be skipped." Users have no way to diagnose this. The docs in Connect-ZtAssessment.md say the private key must be installed in the local cert store but don't say it must be marked as exportable.

   **Fix:** Add to the "Certificate" section in the docs: *The certificate's private key must be marked as exportable (required for SharePoint Online CBA, which uses a temporary PFX export workaround).*

2. **`$pfxBytes` lingers in memory as plaintext.** Minor — PowerShell doesn't provide secure byte wiping — but `$pfxBytes = $null` immediately after `WriteAllBytes` limits the exposure window. Optional hardening.

src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -256,6 +276,30 @@ function Connect-ZtAssessment {
...
 				Add-ZtConnectedService -Service 'Graph'
+				# Resolve the initial domain once here so EXO, S&C, and SPO can reuse it without redundant Graph calls.
+				try {
+					$resolvedOrg = Invoke-ZtGraphRequest -RelativeUri 'organization' -ErrorAction Stop
+					$script:resolvedInitialDomain = $resolvedOrg.verifiedDomains | Where-Object { $_.isInitial } | Select-Object -ExpandProperty name -First 1
+				}
+				catch {
+					Write-Verbose -Message "Failed to resolve initial domain after Graph connect: $($_.Exception.Message)"
+				}
+				# Fallback: try Azure context if Graph did not resolve the initial domain.
```
alexandair: Azure Domain Fallback Runs Before Azure Is Connected — Limited Utility

```powershell
# Fallback: try Azure context if Graph did not resolve the initial domain.
if (-not $script:resolvedInitialDomain -and (Get-Command -Name Get-AzTenant -ErrorAction Ignore)) {
    $azTenantDomains = (Get-AzTenant -ErrorAction Stop).Domains
```

This block executes inside the `'Graph'` case of the `switch`, before the `'Azure'` case processes. In a fresh session, `Get-AzTenant` has no connected account and throws `Az.Accounts\Get-AzTenant: Run Connect-AzAccount to login.` (caught and silently swallowed). The fallback only works when Azure happened to be connected from a prior session — which is exactly the reconnect scenario where `$script:resolvedInitialDomain` was cleared by `Remove-ZtConnectedService`.

This is not a hard bug (it's wrapped in try/catch) but the comment implies a reliable fallback. In practice it's a no-op in new sessions. Consider removing the fallback entirely or moving it to after the `'Azure'` case completes — but that would require a restructuring of the switch. At minimum, the comment should be honest about the limitation.

alexandair (changes requested):

src/powershell/public/Connect-ZtAssessment.ps1
alexandair: **Azure catch block not updated to use `$_.Exception.Message`**

Every other service catch block (AIP, EXO, S&C, SPO) was updated in this PR to use `$_.Exception.Message` for the user-visible "Error details:" line. The Azure catch block was missed:

```powershell
# Azure – still uses $_ (current code, ~line 441)
catch {
    Write-PSFMessage -Message ("Failed to authenticate to Azure: {0}" -f $_) -Level Debug -ErrorRecord $_
    Remove-ZtConnectedService -Service 'Azure'
    Write-Host -Object "   ❌ Failed to connect." -ForegroundColor Yellow
    Write-Host -Object "      Tests requiring Azure will be skipped." -ForegroundColor Yellow
    Write-Host -Object ("       Error details: {0}" -f $_) -ForegroundColor Red   # ← should be $_.Exception.Message
}
```

Using `$_` in a format string calls `.ToString()` on the ErrorRecord, dumping the full stack trace to the console — noisier than intended, inconsistent with every other service.

**Fix:**
```powershell
Write-PSFMessage -Message ("Failed to authenticate to Azure: {0}" -f $_.Exception.Message) -Level Debug -ErrorRecord $_
# ...
Write-Host -Object ("       Error details: {0}" -f $_.Exception.Message) -ForegroundColor Red
```

src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -647,63 +741,60 @@ function Connect-ZtAssessment {
...
+						# Connect-SPOService is loaded via WinPS5 implicit remoting. The -CertificateThumbprint parameter
+						# searches Cert:\CurrentUser\My inside the WinPS5 process which may not see the current session's
+						# cert store. Export the cert to a temp PFX and pass its path + password to bypass store lookup.
+						$tempPfxPath = [System.IO.Path]::Combine([System.IO.Path]::GetTempPath(), [System.IO.Path]::GetRandomFileName() + '.pfx')
+						$tempPassword = [System.Convert]::ToBase64String([System.Security.Cryptography.RandomNumberGenerator]::GetBytes(32))
+						$tempSecurePassword = ConvertTo-SecureString $tempPassword -AsPlainText -Force
+						try {
+							$pfxBytes = $Certificate.Certificate.Export([System.Security.Cryptography.X509Certificates.X509ContentType]::Pfx, $tempPassword)
+							[System.IO.File]::WriteAllBytes($tempPfxPath, $pfxBytes)
+							$pfxBytes = $null  # Limit plaintext private key material in memory
```
alexandair: **Minor hardening — `$tempPassword` plain string not cleared after SPO PFX export**

`$pfxBytes = $null` is correctly zeroed after `WriteAllBytes`. `$tempPassword` (the base64 PFX password string) is not. Since the file is deleted in `finally`, the remaining risk is a memory dump exposing the password that unlocks a now-deleted file — low severity, but inconsistent with the `$pfxBytes` treatment:

```powershell
[System.IO.File]::WriteAllBytes($tempPfxPath, $pfxBytes)
$pfxBytes = $null           # ← already done ✅
$tempPassword = $null       # ← add this for consistency
Connect-SPOService ...
```

src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -647,63 +741,60 @@ function Connect-ZtAssessment {
...
+						}
+						catch [System.Security.Cryptography.CryptographicException] {
+							Write-Host -Object "   ❌ Failed to export the certificate private key for SharePoint Online." -ForegroundColor Red
+							Write-Host -Object "      The certificate's private key must be marked as exportable." -ForegroundColor Yellow
+							Write-Host -Object "      Re-import the certificate with 'Mark this key as exportable' enabled and try again." -ForegroundColor Yellow
+							Write-PSFMessage -Message ("Failed to export certificate private key for SharePoint Online CBA: {0}" -f $_.Exception.Message) -Level Debug -ErrorRecord $_
+							Remove-ZtConnectedService -Service 'SharePointOnline'
+							continue
+						}
+						finally {
```
alexandair, line 805: **Minor — temp PFX delete failure is silently swallowed**

```powershell
finally {
    if (Test-Path $tempPfxPath -ErrorAction SilentlyContinue) {
        Remove-Item $tempPfxPath -Force -ErrorAction SilentlyContinue
        # ← no notification if delete fails
    }
}
```

If `Remove-Item` fails (locked file, permissions), the encrypted PFX stays on disk permanently with no warning. Given the private key is inside it, a `Write-Warning` on failure would at least surface the issue:

```powershell
finally {
    if (Test-Path $tempPfxPath -ErrorAction SilentlyContinue) {
        Remove-Item $tempPfxPath -Force -ErrorAction SilentlyContinue
        if (Test-Path $tempPfxPath -ErrorAction SilentlyContinue) {
            Write-Warning "Failed to delete temporary PFX file '$tempPfxPath'. Please delete it manually."
        }
    }
}
```

alexandair (changes requested):

alexandair (approved):
LGTM

@merill still needs to confirm the required high-privileged Graph permission.

src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -469,23 +467,47 @@ function Connect-ZtAssessment {
...
+				Remove-ZtConnectedService -Service 'AipService'
+			}
+
+			try {
+				Write-PSFMessage -Message "Connecting to Azure Information Protection" -Level Verbose
+				if ($ClientId -and $Certificate) {
+					# Connect-AipService supports app-only CBA via -CertificateThumbprint / -ApplicationId /
+					# -TenantId / -ServicePrincipal (requires AIPService module v1.0.05+ and
+					# Azure Rights Management Services | Application.Read.All permission on the service principal).
+					$null = Connect-AipService `
```
alexandair, line 491: **AIP CBA ignores sovereign cloud — silently breaks USGov/China**

`Connect-AipService` accepts `-AzureEnvironment` (values: `AzureCloud`, `AzureChinaCloud`, `AzureUSGovernment`, `AzureUSGovernment2`, `AzureUSGovernment3`). The new CBA branch in Connect-ZtAssessment.ps1 doesn't pass it, so AIP CBA always targets commercial endpoints regardless of `-Environment`. This is the same class of bug as the SPO-admin-URL-suffix one that was just fixed.

```powershell
$aipParams = @{
    CertificateThumbprint = $Certificate.Certificate.Thumbprint
    ApplicationId         = $ClientId
    TenantId              = $TenantId
    ServicePrincipal      = $true
    ErrorAction           = 'Stop'
}
$aipAzureEnvironment = switch ($Environment) {
    'China'    { 'AzureChinaCloud' }
    'USGov'    { 'AzureUSGovernment' }
    'USGovDoD' { 'AzureUSGovernment2' }   # or 3 — verify against AIP module
    default    { $null }
}
if ($aipAzureEnvironment) { $aipParams.AzureEnvironment = $aipAzureEnvironment }
$null = Connect-AipService @aipParams
```

src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -469,23 +467,47 @@ function Connect-ZtAssessment {
...
-						# but silently fail regardless of permissions. Skip to avoid triggering browser auth.
-						Write-Host -Object "   ⚠️ AipService does not support app-only (certificate) authentication." -ForegroundColor Yellow
-						Write-Host -Object "      AIP tests will be skipped." -ForegroundColor Yellow
-						Remove-ZtConnectedService -Service 'AipService'
-						continue
-					}
-					# Connect-AipService does not support certificate-based authentication via an X509Certificate2
-					# object or app-only service principal flow. Connect using the existing authenticated session;
-					# it will use the Graph/Az context or prompt interactively if needed.
+				$null = Get-AipServiceConfiguration -ErrorAction Stop
```
alexandair: **AIP "already connected" probe via `Get-AipServiceConfiguration` is heavy**

`Get-AipServiceConfiguration` is a real round-trip to the AIP service. Used here purely as an `is-connected?` probe (Connect-ZtAssessment.ps1) — and on a fresh session it always throws (caught). Acceptable, but flagging because every `Connect-ZtAssessment` invocation now incurs this. A cheaper alternative is to track connection state via `$script:ConnectedService -contains 'AipService'` (and only probe under `-Force`). Not blocking.

alexandair (changes requested):

src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -434,18 +467,67 @@ function Connect-ZtAssessment {
...
+			}
+
 			try {
-					Write-PSFMessage -Message "Connecting to Azure Information Protection" -Level Verbose
-					# Connect-AipService does not have parameters for non-interactive auth, so it will use the existing Graph connection context if available, or prompt if not.
-					$null = Connect-AipService -ErrorAction Stop
-					Write-Host -Object "   ✅ Connected" -ForegroundColor Green
-					Add-ZtConnectedService -Service 'AipService'
+				Write-PSFMessage -Message "Connecting to Azure Information Protection" -Level Verbose
+				$aipEnvironmentName = switch ($Environment) {
```
alexandair, line 481: **AIP — `USGovDoD` mapped to wrong sovereign cloud (same class of bug as the SPO `.com` one that was just fixed)**

```powershell
$aipEnvironmentName = switch ($Environment) {
    'China'    { 'AzureChinaCloud' }
    'USGov'    { 'AzureUSGovernment' }
    'USGovDoD' { 'AzureUSGovernment' }   # ← targets GCC High, not DoD
    default    { 'AzureCloud' }
}
```

`Connect-AipService -EnvironmentName` accepts `AzureUSGovernment` (GCC High) **and** `AzureUSGovernment2` (DoD) / `AzureUSGovernment3`. Mapping `USGovDoD → AzureUSGovernment` silently routes DoD tenants through GCC High endpoints. Use `'AzureUSGovernment2'` for DoD (verify against the AIPService module version pinned by the project.).

src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -434,18 +467,67 @@ function Connect-ZtAssessment {
...
+				}
+				catch {
+					Write-PSFMessage -Message "No existing connection to Azure Information Protection found." -Level Debug
+				}
+			}
+			else {
+				Write-PSFMessage -Message "A connection to Azure Information Protection is already tracked by this session." -Level Debug
+			}
+
+			if ($isAipConnected -and -not $Force.IsPresent) {
```
alexandair: **AIP "already connected" path ignores parameter mismatch (inconsistent with Graph/Azure)**

Graph and Azure both reconnect when `ClientId`/`TenantId`/`Certificate` change between calls. AIP only checks "is connected?" — if a user previously connected interactively in the same session and re-runs `Connect-ZtAssessment ... -Certificate`, AIP keeps the delegate session and reports "✅ Already connected", silently bypassing CBA. Workaround today is `-Force`, but this is inconsistent with the other services. Either add a CBA-vs-delegate detection or document the limitation.

src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -113,12 +115,18 @@ function Connect-ZtAssessment {
 	}
 	if ($IgnoreLanguageMode) { $script:IgnoreLanguageMode = $true }
 
+	if ($Certificate -and (-not $ClientId -or -not $TenantId)) {
```
alexandair, line 118: `-UseDeviceCode` and `-Certificate` are not mutually exclusive
Both flow through to Graph (`UseDeviceCode = $true` and `Certificate = ...`) and Azure (`UseDeviceAuthentication = $true` and `ApplicationId + ServicePrincipal`). The SDK calls will fail with opaque errors. Mirror the `-Certificate` guard:

```powershell
if ($Certificate -and $UseDeviceCode) {
    Stop-PSFFunction -Message "-UseDeviceCode cannot be combined with -Certificate (app-only auth is non-interactive)." -EnableException $true -Cmdlet $PSCmdlet
    return
}
```
This should be added after line 121.

alexandair (changes requested):

alexandair (approved):
LGTM

PR #1167, opened 2026-04-18
Add DuckDB Cross Platform Support

billmcilhargey:
Add architecture-specific DuckDB library handling and improve error messages

Add DuckDB support for Linux x64 and Linux arm64 platforms

(moved to dev branch - rather than main branch from earlier PR)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -407,3 +407,6 @@ ZeroTrustReport
 /src/powershell/export
 ZeroTrustReport
 /code-tests/TestResults
+
+# Runtime-generated: copied from arch-specific subdir by Initialize-Dependencies.ps1
+/src/powershell/lib/libduckdb.so
--- a/src/powershell/Initialize-Dependencies.ps1
+++ b/src/powershell/Initialize-Dependencies.ps1
@@ -63,6 +63,49 @@ function Initialize-Dependencies {
 
     Write-Verbose -Message "=== Initialize-Dependencies.ps1 Starting ==="
 
+    #region Resolve architecture-specific native DuckDB library on Linux
+    Write-Host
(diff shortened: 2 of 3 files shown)


kodevza:
@aahmed-spec @alexandair  Tested on Ubuntu with ZeroTrustAssessment 2.3.0.

The package includes duckdb.dll and libduckdb.dylib, but no libduckdb.so. The error shown is about Visual C++ Redistributable.

Bill’s change looks like the right fix and enables me to generate the report on Ubuntu 24.04.

Could you review and merge this into dev? I can provide logs from my side if needed.

kodevza:
@aahmed-spec @merill , one concern with the current DuckDB setup: native libraries are vendored directly into src/powershell/lib, which makes the repo heavier and prevents Dependabot from tracking/updating them properly.

Could we move DuckDB to an external dependency model instead, e.g. NuGet + build/package-time asset copy?

If you’re open to this direction, I can take it forward.

billmcilhargey:
Closing this original implementation because it has been superseded by a more complete RID-based approach in the active development work:

- DuckDB.NET.Data.Full and DuckDB.NET.Bindings.Full are pinned to 1.2.1 with a lock file.
- Managed and native assets are materialized at test/build time rather than committed or copied during module import.
- Native selection supports Windows x64/ARM64, Linux x64/ARM64, and macOS x64/ARM64.
- Windows ARM64 was validated natively with DuckDB v1.2.1.
- The complete PowerShell suite passes: 14,121/14,121.
- CI covers all six OS/architecture combinations.

Thank you for the testing and feedback on this PR; it directly informed the replacement design.

PR #1174, opened 2026-04-20
Fix: Update the fail criteria logic in test 21818

### LINKED ISSUE #1039: ZTA v2.1.0  False‑positive logic in notification settings checks for Entra ID admin roles (Tests 21818 & 21820)
### **Description**

While reviewing ZTA v2.1.0, we noticed that the notification validation logic in tests 21818 and 21820 does not align with how Entra ID role notification settings are designed and are producing false‑positive failures.

### Affected tests
- 21818 – Privileged role activations have monitoring and alerting configured
- 21820 – Activation alert for all privileged role assignments

### **Current behavior**

**Test 21818**

The test fails when default recipients are enabled and no additional recipients are configured:
`
if ($rule.isDefaultRecipientsEnabled -eq $true -and
    $rule.notificationRecipients.Count -eq 0) {
    $passed = $false
    $exitLoop = $true
    break
}
`

**Test 21820**

The test checks only the following rule:
`Notification_Requestor_EndUser_Assignment`

and applies similar logic:

`
if ($isDefaultRecipientsEnabled -eq $true -and
    ([string]::IsNullOrEmpty($notificationRecipients) -or
     $notificationRecipients.Count -eq 0)) {
    $passed = $false
}
`

### **Problem**

This logic assumes that:

 * If default recipients are enabled, additional recipients must also be configured

However, this does not match the Entra ID notification model.
In Entra ID:

- Default recipients (Admin / Requestor / Approver) are valid recipients by themselves
- Additional recipients are optional, used only when notifications should be sent beyond default recipients
- It is fully valid (and common) to:
  + Enable default recipients
  + Leave additional recipients empty

With the current logic:

- A valid configuration is flagged as misconfigured
- This leads to false positives in both tests

### **Expected behavior**

A notification rule should fail only when there is no effective recipient, meaning:

- isDefaultRecipientsEnabled = false
AND
- notificationRecipients is empty

Example of safer logic:

`
if ($rule.isDefaultRecipientsEnabled -eq $false -and
    ($null -eq $rule.notificationRecipients -or
     $rule.notificationRecipients.Count -eq 0)) {
    $passed = $false
}
`

**Additional observation (Test 21820)**

Test 21820 currently validates: `Notification_Requestor_EndUser_AssignmentShow`
However, for activation alert coverage, it would be more appropriate to validate: `Notification_Admin_EndUser_AssignmentShow`

This rule is typically where organizations configure:
- SOC / monitoring mailboxes
- Security operations notifications

Checking only the Requestor rule does not accurately validate monitoring and alerting intent.

Image

praneeth-0000:
Change the Fail scenario logic to fail when isDefaultRecipientsEnabled is false and notificationRecipients is empty.

closes #1039

### DIFF
--- a/src/powershell/tests/Test-Assessment.21818.ps1
+++ b/src/powershell/tests/Test-Assessment.21818.ps1
@@ -123,9 +123,9 @@ ORDER BY rd.displayName;
                     $notificationRules += ($rule | Add-Member -MemberType NoteProperty -Name RoleDisplayName -Value $roleDisplayName -Force -PassThru)
 
                     # TO-DO: When the performance of the API is improved, we can collect all rules and move the check outside the loop to determine if the test passes or fails.
-                    # Check if isDefaultRecipientsEnabled is true and notificationRecipients is an empty array
-                    if ($rule.isDefaultRecipientsEnabled -eq $true -and
-                        $rule.notificationRecipients.Count -eq 0) {
+                    # Check if isDefaultRecipientsEnabled is false and notificationRecipients is an empty array
+                    if ($rule.isDefaultRecipientsEnabled -eq $false -and
+                        ([string]::IsNullOrEmpty($rule.notificationRecipients) -or $rule.notificationRecipients.Count -eq 0)) {
                         $passed = $false
                         $exitLoop = $true
                         break  # Exit inner loop if condition is met
--- a/src/powershell/tests/Test-Assessment.21818.ps1
+++ b/src/powershell/tests/Test-Assessment.21818.ps1
@@ -125,7 +125,7 @@ ORDER BY rd.displayName;
                     # TO-DO: When the performance of the API is improved, we can collect all rules and move the check outside the loop to determine if the test passes or fails.
                     # Check if isDefaultRecipientsEnabled is false and notificationRecipients is an empty array
                     if ($rule.isDefaultRecipientsEnabled -eq $false -and
-                        ([string]::IsNullOrEmpty($rule.notificationRecipients) -or $rule.notificationRecipients.Count -eq 0)) {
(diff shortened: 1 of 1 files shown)


src/powershell/tests/Test-Assessment.21818.ps1
```diff
@@ -123,9 +123,9 @@ ORDER BY rd.displayName;
                     $notificationRules += ($rule | Add-Member -MemberType NoteProperty -Name RoleDisplayName -Value $roleDisplayName -Force -PassThru)
 
                     # TO-DO: When the performance of the API is improved, we can collect all rules and move the check outside the loop to determine if the test passes or fails.
-                    # Check if isDefaultRecipientsEnabled is true and notificationRecipients is an empty array
-                    if ($rule.isDefaultRecipientsEnabled -eq $true -and
-                        $rule.notificationRecipients.Count -eq 0) {
+                    # Check if isDefaultRecipientsEnabled is false and notificationRecipients is an empty array
+                    if ($rule.isDefaultRecipientsEnabled -eq $false -and
+                        ([string]::IsNullOrEmpty($rule.notificationRecipients) -or $rule.notificationRecipients.Count -eq 0)) {
```
alexandair: The `[string]::IsNullOrEmpty()` usage on a collection property should be replaced with `$null -eq` check.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.21818.ps1
praneeth-0000: Hi @alexandair , I've updated the code.

alexandair (approved):
LGTM

SagarSathe (approved):

PR #1186, opened 2026-04-24
Bugfix 1152

### LINKED ISSUE #1152: Binder Error: Could not find key "userprincipalname" in struct Candidate Entries: "uniqueName"
Finding an error when trying to run the identity module:

WARNING: [13:30:25][Wait-ZtTenantDataExport] Export 'ServicePrincipal' failed: Cannot bind argument to parameter 'ArgumentList' because it is an empty array.

Exception: C:\Users\xxx\OneDrive - xxx\General\Documents\PowerShell\Modules\ZeroTrustAssessment\2.2.0\private\export\Export-Database.ps1:115

Line |
 115 |          Invoke-DatabaseQuery -Database $Database -Sql $sql -NonQuery
     |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Exception calling "ExecuteNonQuery" with "0" argument(s): "Binder Error: Could not find key "userprincipalname" in struct  Candidate Entries: "uniqueName""

ashwinikarke:
When the Graph API returned a page with no items (`{ "value": [] }`) for `beta/servicePrincipals, Add-GraphProperty` passed the empty list directly to `Invoke-ZtGraphBatchRequest -ArgumentList`. PowerShell cannot bind an empty array to a mandatory `[object[]]` parameter, causing:

`WARNING: Export 'ServicePrincipal' failed: Cannot bind argument to parameter 'ArgumentList' because it is an empty array.`

This caused the entire ServicePrincipal export to fail.

Fix

Added a guard in Add-GraphProperty that returns early when the results list is empty, skipping the batch call entirely:
```
if (-not $Results -or @($Results).Count -eq 0) {
    return
}
```

### DIFF
--- a/code-tests/commands/Export-ZtGraphEntity.Tests.ps1
+++ b/code-tests/commands/Export-ZtGraphEntity.Tests.ps1
@@ -1,0 +1,117 @@
+Describe "Export-ZtGraphEntity" {
+    # Regression tests for customer bug v2.2.0:
+    #
+    #   Error 1 (WARNING): Export 'ServicePrincipal' failed:
+    #     Cannot bind argument to parameter 'ArgumentList' because it is an empty array.
+    #
+    #   Root cause: When a Graph API page returned { "value": [] }, Add-GraphProperty
+    #   passed the empty $Results.Value directly to Invoke-ZtGraphBatchRequest -ArgumentList.
+    #   PowerShell cannot bind @() to a mandatory [object[]] parameter.
+    #
+    #   Fix: Guard in Add-GraphProperty:
+    #     if (-not $Results -or @($Results).Count -eq 0) { return }
+    #
+    #   Error 2 (Exception): Export-Database.ps1:115
+    #     Binder Error: Could not find key "userprincipalname" in struct
+    #
+    #   Root cause: When all role principals are Service Principals, DuckDB inferred the
+    #   principal struct without userPrincipalName, causing the vwRole SQL to fail.
+    #
+    #   Fix: Model JSON files pre-declare all principal fields as null.
+    #   See Export-Database.Tests.ps1 for Error 2 regression tests.
+
+    BeforeAll {
+        $srcRoot = Join-Path $PSScriptRoot "../../src/powershell"
+        if (-not (Get-Module ZeroTrustAssessment -ErrorAction SilentlyContinue)) {
+            Import-Module (Join-Path $srcRoot "ZeroTrustAssessment.psd1") -Global 3>$null
+            Import-Module (Join-Path $srcRoot "ZeroTrustAssessment.psm1") -Global -Force 3>$null
+        }
+        if (-not (Get-Command Get-MgContext -ErrorAction SilentlyContinue)) {
+            function global:Get-MgContext {}
+        }
+        if (-not (Get-Command Invoke-MgGraphRequest -ErrorAction SilentlyContinue)) {
+            function global:Invoke-MgGraphRequest { param($Method, $Uri, $OutputType) }
+        }
+    }
+
+    Context "Error 1 reproduction — empty page triggers ArgumentList error (old behaviour)" {
+        # Invoke-ZtGraphBatchRequest declares ArgumentList as mandatory [object[]].
+        # Before the fix, Add-GraphProperty passed $Results.Value directly with no guard,
+        # so an empty page caused the customer error.
+
+        It "Invoke-ZtGraphBatchRequest throws when ArgumentList is empty" {
+            { Invoke-ZtGraphBatchRequest -Path 'servicePrincipals/{0}/oauth2PermissionGrants' -ArgumentList @() } |
+                Should -Throw
+        }
+
+        It "The error message matches the customer warning ('ArgumentList')" {
+            $msg = $null
+            try {
+                Invoke-ZtGraphBatchRequest -Path 'servicePrincipals/{0}/oauth2PermissionGrants' -ArgumentList @()
+            }
+            catch { $msg = $_.Exception.Message }
+            $msg | Should -Match 'ArgumentList'
+        }
+    }
+
+    Context "Error 1 fix — guard in Add-GraphProperty skips batch when page is empty" {
+        BeforeAll {
+            $script:exportPath = Join-Path $env:TEMP "zt-test-graphentity-$(Get-Random)"
+            New-Item -ItemType Directory -Path $script:exportPath -Force | Out-Null
+        }
+
+        AfterAll {
+            Remove-Item $script:exportPath -Recurse -Force -ErrorAction SilentlyContinue
+        }
+
+        BeforeEach {
+            Mock -ModuleName ZeroTrustAssessment Get-ZtConfig            { return $false }
+            Mock -ModuleName ZeroTrustAssessment Set-ZtConfig            {}
+            Mock -ModuleName ZeroTrustAssessment Update-ZtProgressState  {}
+            Mock -ModuleName ZeroTrustAssessment Get-PSFConfigValue      { return 1073741824 }
+        }
+
+        It "Does not throw when Graph API returns an empty page" {
+            # Invoke-ZtRetry returns { "value": [] } — same as the customer tenant.
+            # The guard must return early without calling Invoke-ZtGraphBatchRequest.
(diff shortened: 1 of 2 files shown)


src/powershell/private/export/Export-ZtGraphEntity.ps1
```diff
@@ -130,6 +130,14 @@ function Export-ZtGraphEntity {
 		Write-PSFMessage -Message "Adding {0} to {1}" -StringValues $PropertyName, $Name -Tag Graph
 		Update-ZtProgressState -WorkerId $Name -WorkerName $Name -WorkerStatus 'Running' -WorkerDetail "Batch: $PropertyName"
 
+		# Guard against empty pages: Invoke-ZtGraphBatchRequest requires a non-empty ArgumentList.
+		# Graph API can return a page with an empty 'value' array (e.g. with $expand queries),
+		# which would cause a PowerShell parameter-binding error for the mandatory [object[]] param.
+		if (-not $Results -or @($Results).Count -eq 0) {
```
alexandair: Redundant condition in the guard

```powershell
if (-not $Results -or @($Results).Count -eq 0) {
```

`-not @()` evaluates to `$true` in PowerShell — an empty array is falsy. So `@($Results).Count -eq 0` is logically redundant. The `@()` re-wrapping is also unnecessary since `$Results` is already an array coming from `$items = $Results.Value`. Simplified form:

```powershell
if (-not $Results) {
```

The existing condition is harmless but adds noise.

alexandair (changes requested):

src/powershell/private/export/Export-ZtGraphEntity.ps1
```diff
@@ -130,6 +130,14 @@ function Export-ZtGraphEntity {
 		Write-PSFMessage -Message "Adding {0} to {1}" -StringValues $PropertyName, $Name -Tag Graph
 		Update-ZtProgressState -WorkerId $Name -WorkerName $Name -WorkerStatus 'Running' -WorkerDetail "Batch: $PropertyName"
 
+		# Guard against empty pages: Invoke-ZtGraphBatchRequest requires a non-empty ArgumentList.
+		# Graph API can return a page with an empty 'value' array (e.g. with $expand queries),
```
alexandair:
```powershell
# Guard against empty pages: Invoke-ZtGraphBatchRequest requires a non-empty ArgumentList.
# Graph API can return a page with an empty 'value' array (e.g. with $expand queries),
```

The parenthetical "(e.g. with $expand queries)" is incorrect. `Export-ZtGraphEntity` doesn't use `$expand` — it uses separate batch requests for `$RelatedPropertyNames`. The empty page was caused by standard paging (tenant has 0 service principals, or the last page of a paginated response returns 0 items). The comment should say something like "(e.g. when the tenant has no items, or on the last page of a paginated response)".

code-tests/commands/Export-ZtGraphEntity.Tests.ps1
```diff
@@ -0,0 +1,84 @@
...
+            Mock -ModuleName ZeroTrustAssessment Invoke-ZtGraphBatchRequest {}
+
+            Export-ZtGraphEntity -Name 'ServicePrincipal' -Uri 'beta/servicePrincipals' `
+                -QueryString '$top=999' -RelatedPropertyNames @('oauth2PermissionGrants') `
+                -ExportPath $script:exportPath
+
+            Should -Invoke -ModuleName ZeroTrustAssessment -CommandName Invoke-ZtGraphBatchRequest -Times 0 -Exactly
+        }
+
+        It "Invoke-ZtGraphBatchRequest is called once when the page has items" {
```
alexandair, line 89: **Test 3 has a conceptual flaw — the second mock branch never executes**

```powershell
It "Invoke-ZtGraphBatchRequest is called once when the page has items" {
    $script:call = 0
    Mock -ModuleName ZeroTrustAssessment Invoke-ZtRetry {
        $script:call++
        if ($script:call -eq 1) { return @{ value = @(@{ id = 'sp-1'; displayName = 'TestSP' }) } }
        return @{ value = @() }   # ← THIS NEVER EXECUTES
    }
```

The first mock return is `@{ value = @(items) }` — a hashtable with **no `@odata.nextLink` key**. In the production code:

```powershell
$actualUri = $results.'@odata.nextLink'   # → $null
if (-not $actualUri) { break }            # → breaks immediately
```

The `do-while` loop exits after the first page. `Invoke-ZtRetry` is called exactly once; `$script:call` reaches 1; the `return @{ value = @() }` branch is dead code. The test passes for the right reason (batch is called once for the non-empty page), but the stated intent — "verify the guard does not suppress normal non-empty pages while correctly skipping empty second pages" — is **not tested**.

To actually test two-page pagination, the first mock return needs `@odata.nextLink`:

```powershell
if ($script:call -eq 1) {
    return @{
        'value'           = @(@{ id = 'sp-1'; displayName = 'TestSP' })
        '@odata.nextLink' = 'https://graph.microsoft.com/beta/servicePrincipals?$skiptoken=abc'
    }
}
return @{ value = @() }   # now this executes on call 2
```

Without this, `Invoke-ZtRetry` only needs one mock and the `$script:call` counter/second-branch is dead complexity.

code-tests/commands/Export-ZtGraphEntity.Tests.ps1
```diff
@@ -0,0 +1,84 @@
...
+    BeforeAll {
+        $srcRoot = Join-Path $PSScriptRoot "../../src/powershell"
+        if (-not (Get-Module ZeroTrustAssessment -ErrorAction SilentlyContinue)) {
+            Import-Module (Join-Path $srcRoot "ZeroTrustAssessment.psd1") -Global 3>$null
+            Import-Module (Join-Path $srcRoot "ZeroTrustAssessment.psm1") -Global -Force 3>$null
+        }
+        if (-not (Get-Command Get-MgContext -ErrorAction SilentlyContinue)) {
+            function global:Get-MgContext {}
+        }
+        if (-not (Get-Command Invoke-MgGraphRequest -ErrorAction SilentlyContinue)) {
```
alexandair: **Unused `Invoke-MgGraphRequest` global stub**

```powershell
if (-not (Get-Command Invoke-MgGraphRequest -ErrorAction SilentlyContinue)) {
    function global:Invoke-MgGraphRequest { param($Method, $Uri, $OutputType) }
}
```

Since all three tests mock `Invoke-ZtRetry` directly, the inner scriptblock `{ Invoke-MgGraphRequest ... }` never executes. The `Invoke-MgGraphRequest` stub in `BeforeAll` is never called. It can be removed from this file without any impact; its presence implies the tests exercise a code path they don't.

code-tests/commands/Export-ZtGraphEntity.Tests.ps1
```diff
@@ -0,0 +1,84 @@
...
+            }
+            Mock -ModuleName ZeroTrustAssessment Invoke-ZtGraphBatchRequest { return @() }
+
+            Export-ZtGraphEntity -Name 'ServicePrincipal' -Uri 'beta/servicePrincipals' `
+                -QueryString '$top=999' -RelatedPropertyNames @('oauth2PermissionGrants') `
+                -ExportPath $script:exportPath
+
+            Should -Invoke -ModuleName ZeroTrustAssessment -CommandName Invoke-ZtGraphBatchRequest -Times 1 -Exactly
+        }
+    }
```
alexandair, line 114: **Missing `$null` coverage**

The fix handles both `$null` and `@()` inputs, but there's no test for `$Results = $null` (as distinct from `$Results = @()`). These are different PowerShell values and worth testing separately if the goal is a comprehensive regression suite:

```powershell
It "Does not throw when Invoke-ZtRetry returns null" {
    Mock -ModuleName ZeroTrustAssessment Invoke-ZtRetry { return $null }
    ...
}
```

alexandair (changes requested):

The fix itself is OK. The test for Test 3 should be corrected — either add `@odata.nextLink` to make it actually test two-page pagination, or simplify it to remove the counter/second-branch dead code and just verify the single-page case with items.

alexandair (approved):
LGTM

SagarSathe (approved):

PR #1190, opened 2026-04-27
bugfix/issue-1183: Use Where-Object instead of index access for targets

### LINKED ISSUE #1183: 21790 is using indexed values to filter user and application targets
21790 is using indexed values to filter user and application targets, we should filter with Where-Object instead

sandeepjha000:
Use Where-Object instead of index access for targets
Fixes #1183

### DIFF
--- a/src/powershell/tests/Test-Assessment.21790.ps1
+++ b/src/powershell/tests/Test-Assessment.21790.ps1
@@ -35,10 +35,10 @@ function Test-Assessment-21790 {
             [object]$TargetConfig
         )
 
-        if ($TargetConfig.targets | Where-Object { $_.target -eq "AllUsers" }) {
+        if ($TargetConfig.targets.target -contains "AllUsers") {
             return "All users"
         }
-        elseif ($TargetConfig.targets | Where-Object { $_.target -eq "AllApplications" }) {
+        elseif ($TargetConfig.targets.target -contains "AllApplications") {
             return "All external applications"
         }
         else {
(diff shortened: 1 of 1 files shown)


src/powershell/tests/Test-Assessment.21790.ps1
```diff
@@ -69,15 +69,15 @@ function Test-Assessment-21790 {
 
     # Evaluate B2B Collaboration outbound settings
```
alexandair, line 70: **`Where-Object` in boolean context is non-idiomatic**

Lines changed in the evaluation block:
```powershell
($result.b2bCollaborationOutbound.usersAndGroups.targets | Where-Object { $_.target -eq "AllUsers" }) -and
```

`Where-Object` returns a `PSCustomObject` (or array of them), not `$true`/`$false`. PowerShell's boolean coercion makes this work — a non-null object is truthy — but it is not obvious to a reader or future maintainer that this is a boolean check.

**The idiomatic PowerShell pattern for this is member enumeration + `-contains`:**
```powershell
($result.b2bCollaborationOutbound.usersAndGroups.targets.target -contains "AllUsers") -and
```

PowerShell's member enumeration automatically projects the `.target` property from every element of the array, and `-contains` is a proper boolean operator. This is:
- More readable (clearly returns a bool)
- Simpler (no pipeline)
- More consistent with how PowerShell handles null gracefully (if `targets` is null, `.target` returns null, and `-contains` is still false)

The same applies to the `Get-TargetDescription` helper function:
```powershell
# PR version
if ($TargetConfig.targets | Where-Object { $_.target -eq "AllUsers" }) {
# Better
if ($TargetConfig.targets.target -contains "AllUsers") {
```

alexandair (changes requested):

Could you also check the following discovered during review and address it in this or a separate PR?

**Bug — 🔴 CRITICAL (pre-existing): Missing `isServiceDefault` check**

**The spec's pass condition explicitly states:**
> **Pass**: The default cross-tenant access policy has `isServiceDefault: false` AND both `b2bCollaborationOutbound` and `b2bDirectConnectOutbound` sections have `accessType: "blocked"` ...

**The code has no check for `isServiceDefault`:**
```powershell
# MISSING from the test:
$result.isServiceDefault -eq $false
```

When `isServiceDefault` is `$true`, the tenant is using the Microsoft-provided default policy, which may coincidentally have `accessType: "blocked"` values — but the tenant has not **explicitly configured** it. The spec considers this a **Fail**. The test can currently produce a false positive pass for tenants that have never touched this setting.

**Fix needed (in addition to this PR):**
```powershell
$b2bCollaborationOutbound = $null -ne $result -and
    $result.isServiceDefault -eq $false -and
    $result.b2bCollaborationOutbound.usersAndGroups.accessType -eq "blocked" -and
    ($result.b2bCollaborationOutbound.usersAndGroups.targets.target -contains "AllUsers") -and
    $result.b2bCollaborationOutbound.applications.accessType -eq "blocked" -and
    ($result.b2bCollaborationOutbound.applications.targets.target -contains "AllApplications")
```

alexandair (approved):
LGTM

PR #1191, opened 2026-04-27
Bugfix/issue 1176: Update application ownership check to exclude servicePrincipalType: 'ManagedIdentities'

### LINKED ISSUE #1176: Managed Identities should be excluded from Enterprise applications with high privilege Microsoft Graph API permissions have owners check
Managed Identities should be excluded as it doesn't appear to be possible to add an Owner.

sandeepjha000:
Update application ownership check to exclude servicePrincipalType: 'ManagedIdentities'

Fixes #1176

### DIFF
--- a/src/powershell/private/tests-shared/Get-ApplicationsWithInsufficientOwners.ps1
+++ b/src/powershell/private/tests-shared/Get-ApplicationsWithInsufficientOwners.ps1
@@ -22,8 +22,9 @@ function Get-ApplicationsWithInsufficientOwners {
     $allApps = Get-ApplicationsWithPermissions -Database $Database
 
     # Filter by privilege level and owner count using Where-Object (more efficient)
+    # Only include Application type
     $filteredApps = $allApps | Where-Object {
-        ($PrivilegeLevel -contains $_.Risk) -and ($_.OwnerCount -lt 2)
+        ($PrivilegeLevel -contains $_.Risk) -and ($_.OwnerCount -lt 2) -and ($_.servicePrincipalType -eq 'Application')
     }
 
     Write-PSFMessage "Filtered to $($filteredApps.Count) applications with < 2 owners matching privilege levels: $($PrivilegeLevel -join ', ')" -Level Verbose
--- a/src/powershell/private/tests-shared/Get-ApplicationsWithInsufficientOwners.ps1
+++ b/src/powershell/private/tests-shared/Get-ApplicationsWithInsufficientOwners.ps1
@@ -18,13 +18,12 @@ function Get-ApplicationsWithInsufficientOwners {
         [string[]]$PrivilegeLevel
     )
 
-    # Get all apps with permissions
-    $allApps = Get-ApplicationsWithPermissions -Database $Database
+    # Get apps with permissions, pre-filtered to Application type only to avoid enriching non-ownable types (e.g. Managed Identities)
+    $allApps = Get-ApplicationsWithPermissions -Database $Database -ServicePrincipalType 'Application'
 
-    # Filter by privilege level and owner count using Where-Object (more efficient)
-    # Only include Application type
+    # Filter the retrieved applications by privilege level and owner count
     $filteredApps = $allApps | Where-Object {
-        ($PrivilegeLevel -contains $_.Risk) -and ($_.OwnerCount -lt 2) -and ($_.servicePrincipalType -eq 'Application')
+        ($PrivilegeLevel -contains $_.Risk) -and ($_.OwnerCount -lt 2)
     }
 
     Write-PSFMessage "Filtered to $($filteredApps.Count) applications with < 2 owners matching privilege levels: $($PrivilegeLevel -join ', ')" -Level Verbose
--- a/src/powershell/private/tests-shared/Get-ApplicationsWithInsufficientOwners.ps1
+++ b/src/powershell/private/tests-shared/Get-ApplicationsWithInsufficientOwners.ps1
@@ -18,13 +18,11 @@ function Get-ApplicationsWithInsufficientOwners {
         [string[]]$PrivilegeLevel
     )
 
-    # Get apps with permissions, pre-filtered to Application type only to avoid enriching non-ownable types (e.g. Managed Identities)
-    $allApps = Get-ApplicationsWithPermissions -Database $Database -ServicePrincipalType 'Application'
+    # Get all apps with permissions, excluding Managed Identities before enrichment as owners cannot be assigned to them
+    $allApps = Get-ApplicationsWithPermissions -Database $Database -ExcludeServicePrincipalType 'ManagedIdentity'
 
-    # Filter the retrieved applications by privilege level and owner count
-    $filteredApps = $allApps | Where-Object {
-        ($PrivilegeLevel -contains $_.Risk) -and ($_.OwnerCount -lt 2)
-    }
+    # Filter by privilege level and owner count
+    $filteredApps = $allApps | Where-Object {($PrivilegeLevel -contains $_.Risk) -and ($_.OwnerCount -lt 2)}
 
     Write-PSFMessage "Filtered to $($filteredApps.Count) applications with < 2 owners matching privilege levels: $($PrivilegeLevel -join ', ')" -Level Verbose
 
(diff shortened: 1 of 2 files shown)


src/powershell/private/tests-shared/Get-ApplicationsWithPermissions.ps1
```diff
@@ -12,15 +12,18 @@ function Get-ApplicationsWithPermissions {
...
     # Query ServicePrincipal objects with permissions
     # Used by tests 21770, 24518, and 21867
     $sql = @"
 select sp.id, sp.appId, sp.displayName, sp.appOwnerOrganizationId, sp.publisherName,
 spsi.lastSignInActivity.lastSignInDateTime,
-sp.owners, sp.signInAudience
+sp.owners, sp.signInAudience, sp.servicePrincipalType
 from main.ServicePrincipal sp
     left join main.ServicePrincipalSignIn spsi on spsi.appId = sp.appId
 where sp.id in
```
alexandair: Pre-filter is post-DB, not in SQL (minor — already raised by Copilot reviewer)
The filter runs in PowerShell *after* the full result set is materialized. The DB query still returns all `ServicePrincipal` rows including managed identities. On large tenants, pushing this into SQL avoids loading rows that will be discarded:

```sql
where sp.servicePrincipalType in ('Application', ...)
  and (sp.id in (...) or sp.id in (...))
```

This is the change the existing comment ("Filter the retrieved applications…") would actually justify. Not blocking, but worth doing while the file is open. The two prior Copilot review threads on this file are already resolved, so I'd raise it as a follow-up rather than block.

src/powershell/private/tests-shared/Get-ApplicationsWithPermissions.ps1
```diff
@@ -12,15 +12,18 @@ function Get-ApplicationsWithPermissions {
     [CmdletBinding()]
     param(
         [Parameter(Mandatory = $true)]
-        $Database
+        $Database,
+
+        [Parameter()]
+        [string[]]$ServicePrincipalType
```
alexandair: Missing `ValidateSet` on the new parameter

```powershell
[Parameter()]
[string[]]$ServicePrincipalType
```

Graph's `servicePrincipalType` is a closed enum (`Application`, `ManagedIdentity`, `Legacy`, `SocialIdp`, `LegacyUpdatedApp`). Adding `[ValidateSet(...)]` makes typos (e.g. `'application'` lowercase, which would silently match nothing because the comparison is case-sensitive on the DB column value) fail fast.

alexandair (changes requested):

Out-of-scope for this PR but worth noting: Test-21770 also targets "applications"
Test-Assessment.21770.ps1 calls `Get-ApplicationsWithPermissions -Database $Database` with no type filter. The 21770 spec also targets enterprise applications via the `/applications` endpoint, so managed identities are likely false positives there too. Consider opening a follow-up issue to apply the same filter to 21770 (out of scope for this PR per the issue title).

src/powershell/private/tests-shared/Get-ApplicationsWithInsufficientOwners.ps1
```diff
@@ -22,8 +22,9 @@ function Get-ApplicationsWithInsufficientOwners {
     $allApps = Get-ApplicationsWithPermissions -Database $Database
 
     # Filter by privilege level and owner count using Where-Object (more efficient)
+    # Only include Application type
     $filteredApps = $allApps | Where-Object {
-        ($PrivilegeLevel -contains $_.Risk) -and ($_.OwnerCount -lt 2)
+        ($PrivilegeLevel -contains $_.Risk) -and ($_.OwnerCount -lt 2) -and ($_.servicePrincipalType -eq 'Application')
     }
```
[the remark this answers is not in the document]
sandeepjha000: This is Not applicable anymore as we decided to only exclude Managed Identities rather than filtering with servicePrincipalType = 'Application'

src/powershell/private/tests-shared/Get-ApplicationsWithInsufficientOwners.ps1
```diff
@@ -22,8 +22,9 @@ function Get-ApplicationsWithInsufficientOwners {
     $allApps = Get-ApplicationsWithPermissions -Database $Database
 
     # Filter by privilege level and owner count using Where-Object (more efficient)
+    # Only include Application type
```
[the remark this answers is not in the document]
sandeepjha000: comment updated

src/powershell/private/tests-shared/Get-ApplicationsWithPermissions.ps1
sandeepjha000: Updated comment based on code changes, this is fixed.

src/powershell/private/tests-shared/Get-ApplicationsWithPermissions.ps1
sandeepjha000: the parameter name has changed now but I have added validationSet based on MS doc: https://learn.microsoft.com/en-us/graph/api/resources/serviceprincipal?view=graph-rest-1.0

src/powershell/private/tests-shared/Get-ApplicationsWithPermissions.ps1
```diff
@@ -51,6 +55,12 @@ order by spsi.lastSignInActivity.lastSignInDateTime
         return @()
     }
 
+    # Exclude specified service principal types before enrichment to avoid unnecessary per-item DB calls
+    if ($ExcludeServicePrincipalType) {
+        $results = $results | Where-Object { $ExcludeServicePrincipalType -notcontains $_.servicePrincipalType }
+        Write-PSFMessage "Excluded $($ExcludeServicePrincipalType -join ', ') type(s), $($results.Count) service principals remaining" -Level Verbose
```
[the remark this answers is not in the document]
sandeepjha000: fixed in commit [e31e32a](https://github.com/microsoft/zerotrustassessment/pull/1191/commits/e31e32a63e001cff24d3ac5d848c89f9e4c20e8b)

src/powershell/private/tests-shared/Get-ApplicationsWithInsufficientOwners.ps1
```diff
@@ -18,13 +18,11 @@ function Get-ApplicationsWithInsufficientOwners {
...
-    $allApps = Get-ApplicationsWithPermissions -Database $Database
+    # Get all apps with permissions, excluding Managed Identities before enrichment as owners cannot be assigned to them
+    $allApps = Get-ApplicationsWithPermissions -Database $Database -ExcludeServicePrincipalType 'ManagedIdentity'
 
-    # Filter by privilege level and owner count using Where-Object (more efficient)
-    $filteredApps = $allApps | Where-Object {
-        ($PrivilegeLevel -contains $_.Risk) -and ($_.OwnerCount -lt 2)
-    }
+    # Filter by privilege level and owner count
+    $filteredApps = $allApps | Where-Object {($PrivilegeLevel -contains $_.Risk) -and ($_.OwnerCount -lt 2)}
```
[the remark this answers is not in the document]
sandeepjha000: fixed in commit [e31e32a](https://github.com/microsoft/zerotrustassessment/pull/1191/commits/e31e32a63e001cff24d3ac5d848c89f9e4c20e8b)

src/powershell/private/tests-shared/Get-ApplicationsWithPermissions.ps1
```diff
@@ -51,6 +55,12 @@ order by spsi.lastSignInActivity.lastSignInDateTime
         return @()
     }
 
+    # Exclude specified service principal types before enrichment to avoid unnecessary per-item DB calls
```
alexandair: **SQL-level filtering is more efficient than PowerShell-level**

The fix adds `sp.servicePrincipalType` to the SQL `SELECT` and then filters in PowerShell after fetching. For large tenants with many managed identities, a SQL `WHERE sp.servicePrincipalType NOT IN (...)` clause would avoid fetching those rows from DuckDB entirely. The current approach is still significantly better than before (it filters before per-item enrichment calls, as the code comment correctly notes), but the SQL approach would be the cleanest.

alexandair (changes requested):

src/powershell/private/tests-shared/Get-ApplicationsWithPermissions.ps1
sandeepjha000: fixed

alexandair (approved):
LGTM

SagarSathe (approved):

PR #1197, opened 2026-05-04
Add Infrastructure Pillar to ZeroTrustAssessment

### LINKED ISSUE #1153: Feature - Add Infrastructure Pillar
Issue created to track Infrastructure pillar creation

praneeth-0000:


src/powershell/private/tests/Invoke-ZtTests.ps1
```diff
@@ -53,7 +53,7 @@
 		[string[]]
 		$Tests,
 
-		[ValidateSet('All', 'Identity', 'Devices', 'Network', 'Data')]
+		[ValidateSet('All', 'Identity', 'Devices', 'Network', 'Data', 'Infrastructure')]
```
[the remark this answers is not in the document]
praneeth-0000: Not needed

src/powershell/private/core/ConvertTo-ZtMarkdown.ps1
```diff
@@ -0,0 +1,73 @@
...
+	# Convert ordered list items with sequential numbering
+	$liCounter = [ref]0
+	$md = [regex]::Replace($md, '<li[^>]*>', {
+		$liCounter.Value++
+		"`n$($liCounter.Value). "
+	})
+
+	$md = $md -replace '</li>', ''
+	$md = $md -replace '<[ou]l[^>]*>', ''
+	$md = $md -replace '</[ou]l>', ''
```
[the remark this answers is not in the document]
praneeth-0000: resolved

src/report/src/components/test-table/data-table.tsx
```diff
@@ -92,6 +92,13 @@ export function DataTable<TData extends Test, TValue>({
         return data;
     }, [data, pillar]);
 
+    // Default to High risk filter for Infrastructure pillar
+    React.useEffect(() => {
+        if (pillar === "Infrastructure" && pillarFilteredData.some(item => item.TestRisk === "High")) {
+            setSelectedRisks(["High"]);
+        }
+    }, [pillar, pillarFilteredData]);
+
```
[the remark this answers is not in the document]
praneeth-0000: As per discussion with Merill, We need to show high risk tests by default whenever user views the infrastructure pillar.

src/powershell/tests/Test-Assessment.50001.ps1
```diff
@@ -0,0 +1,736 @@
...
+        $recName = $rec.Name
+        $inSS    = $secureScoreGroupHash.ContainsKey($recName)
+
+        # ── SECURE SCORE BACKED (left-only or both intersection) ─────────────────────
+        if ($inSS) {
+            $rows     = $secureScoreGroupHash[$recName].Group
+            $firstRow = $rows[0]
+            $testId   = $recName
+            $title    = $firstRow.recommendationDisplayName
+            $risk     = $firstRow.severity
```
alexandair: **MDC severity `'None'` or `'Informational'` will crash `Add-ZtTestResultDetail`**

```powershell
$risk = $firstRow.severity
```

This value is passed directly as `Risk = $risk` to `Add-ZtTestResultDetail`, which has `[ValidateSet('High', 'Medium', 'Low')]` on its `$Risk` parameter.

MDC's `metadata.severity` field can return `'None'` (for assessments without a severity assignment) and potentially `'Informational'`. Neither is in the ValidateSet. PowerShell validates parameter values before the function body runs, so this will throw a hard exception at runtime for any assessment with a non-standard severity, crashing that loop iteration — and potentially silently dropping all subsequent recommendations if the error isn't caught.

The same bug exists in the MCSB-only path at line ~566:
```powershell
$risk = $firstRow.severity
```

**Fix:**
```powershell
$risk = switch ($firstRow.severity) {
    'High'   { 'High' }
    'Medium' { 'Medium' }
    'Low'    { 'Low' }
    default  { 'Medium' }  # Treat None/Informational as Medium
}
```

src/powershell/private/core/Add-ZtTestResultDetail.ps1
```diff
@@ -198,6 +210,10 @@ function Add-ZtTestResultDetail {
 		$getZtTestStatusParams.CustomStatus = $CustomStatus
 	}
 
+	# Resolve pillar: prefer testMeta, fall back to explicit -Pillar parameter (for dynamic tests)
+	$resolvedPillar = if ($testMeta.Pillar) { $testMeta.Pillar } elseif ($Pillar) { $Pillar } else { $null }
```
alexandair: **`$resolvedPillar` in `Add-ZtTestResultDetail` prefers `$testMeta` over explicit `-Pillar` param — wrong precedence order**


```powershell
# Resolve pillar: prefer testMeta, fall back to explicit -Pillar parameter (for dynamic tests)
$resolvedPillar = if ($testMeta.Pillar) { $testMeta.Pillar } elseif ($Pillar) { $Pillar } else { $null }
```

The comment says `-Pillar` is a fallback for dynamic tests, yet the code always prefers `$testMeta.Pillar`. The `Category` and `Title` params both use `$PSBoundParameters.ContainsKey()` to let explicit caller values win. `Pillar` should follow the same pattern:

```powershell
$resolvedPillar = if ($PSBoundParameters.ContainsKey('Pillar')) { $Pillar }
                  elseif ($testMeta.Pillar) { $testMeta.Pillar }
                  else { $null }
```

For test 50001 this is harmless (both values are `'Infrastructure'`), but the current logic means a dynamic test can never override the pillar set by the `[ZtTest()]` attribute, which contradicts the design intent of the new parameter.

src/powershell/private/core/Add-ZtTestResultDetail.ps1
```diff
@@ -198,6 +210,10 @@ function Add-ZtTestResultDetail {
...
+	$resolvedSfiPillar = if ($testMeta.SfiPillar) { $testMeta.SfiPillar } else { $null }
```
alexandair: **`$resolvedSfiPillar` has no way to be overridden by callers**

Add-ZtTestResultDetail.ps1

```powershell
$resolvedSfiPillar = if ($testMeta.SfiPillar) { $testMeta.SfiPillar } else { $null }
```

If the dynamic test needs per-recommendation SFI pillar values (e.g., future tests), there is no `-SfiPillar` parameter to pass it through. This is a design gap opened by this PR. Not blocking for 50001 but worth noting before this pattern is reused.

alexandair (changes requested):
@praneeth-0000 Just a couple of findings.

src/powershell/tests/Test-Assessment.50001.ps1
praneeth-0000: Good catch, Aleks. This is resolved.

src/powershell/private/core/Add-ZtTestResultDetail.ps1
praneeth-0000: This is fixed.

src/powershell/private/core/Add-ZtTestResultDetail.ps1
praneeth-0000: As of now, SFI pillar need not be passed dynamically. Hence, I made this static as a fix.

alexandair (changes requested):
Here is what I found after analyzing all 5 helper functions in `Test-Assessment.50001.ps1`:

---

`Get-NormalizedRisk`, `Get-DescriptionMarkdown`, `Get-ColumnFlags`, and `Get-McsbCells` are all correct. All bugs are in `Get-RowLinks` (lines 98–105).

1. **Fix `SubMd` — blank link text when `subscriptionName` is empty** (*Bug 1*)  
   If the left-outer join on `resourcecontainers` misses (subscription not in the container list), `subscriptionName` is `''`. The current code produces `[](url)` — blank link text.  
   Fix: fall back to `subscriptionId` as display text when `subscriptionName` is empty.

2. **Fix `ResMd` — blank link text on per-row empty `resourceName`** (*Bug 2*)  
   `Get-ColumnFlags` sets `$showResource = $true` if *any* row has a `resourceName`. Rows in the same recommendation that lack a name produce `[](https://portal.azure.com/#resource{id})` — blank, and a broken URL if `resourceId` is also empty.  
   Fix: fall back to `resourceId` (or a literal `'(resource)'`) as display text when `resourceName` is empty.

3. **Fix `PortalLinkMd` — MCSB-only rows may produce a relative portal URL** (*Bug 3*)  
   The Secure Score KQL (line ~218–220) applies an explicit `case()` to prefix any non-`https://` value before storing `azurePortalRecommendationLink`. The MCSB KQL (line ~307) does **not** — it uses `tostring(properties1.links.azurePortal)` verbatim. For MCSB-only rows, a link like `"portal.azure.com/#blade/..."` renders as a relative URL in the report and breaks navigation.  
   Fix: in `Get-RowLinks`, before building the `[View recommendation](...)` string, ensure the link starts with `https://`; prepend it otherwise.

src/powershell/private/core/ConvertTo-ZtMarkdown.ps1
```diff
@@ -0,0 +1,110 @@
...
+	# (?:...) is a non-capturing group so $1 refers to the text content, not the tag name
+	$md = $md -replace '<(?:b|strong)[^>]*>([^<]*)</(?:b|strong)>', '**$1**'
+
+	# <(?:i|em)[^>]*>([^<]*)</(?:i|em)> — italic tags wrapping plain text; group 1 = content
+	$md = $md -replace '<(?:i|em)[^>]*>([^<]*)</(?:i|em)>', '*$1*'
+
+	# <[^>]+> — any remaining HTML tag: one or more non-'>' characters between angle brackets
+	$md = $md -replace '<[^>]+>', ''
+
+	# Decode HTML entities
```
alexandair: **`ConvertTo-ZtMarkdown`: only 6 HTML entities decoded**

The manual replacement chain covers only `&amp;`, `&lt;`, `&gt;`, `&quot;`, `&#39;`, `&nbsp;`. MDC descriptions in practice contain `&mdash;`, `&ndash;`, `&lsquo;`, `&rsquo;`, `&#8220;`, `&#8221;`, and many numeric entities. These pass through as raw entity strings in the rendered report.

Fix: replace the six manual replacements with a single call to `[System.Net.WebUtility]::HtmlDecode($md)`, which handles the full HTML entity set. The class is available in all supported .NET versions.

src/powershell/private/core/ConvertTo-ZtMarkdown.ps1
```diff
@@ -0,0 +1,110 @@
...
+	# (https?://[^\s<>"\[\]()]+?)  — captures the URL lazily (stops before whitespace or special chars)
+	# ([.,;]?)                     — optionally captures a trailing punctuation character
+	# (?=\s|$)                     — lookahead: must be followed by whitespace or end of string
+	$md = [regex]::Replace($md, '(?<!\()(https?://[^\s<>"\[\]()]+?)([.,;]?)(?=\s|$)', '[${1}](${1})${2}')
+
+	# Break inline numbered steps onto separate lines: "text 2. Word" → "text\n2. Word"
+	# (?<=\S)      — positive lookbehind: preceded by a non-whitespace character (mid-sentence)
+	# (\d{1,2})\. — captures 1–2 digit number followed by a literal dot
+	# ([A-Z])      — lookahead-style capture: next word starts with uppercase (new sentence/step)
+	$md = [regex]::Replace($md, '(?<=\S) (\d{1,2})\. ([A-Z])', "`n" + '$1. $2')
```
alexandair: **`ConvertTo-ZtMarkdown`: numbered-step split may break legitimate prose**

```powershell
$md = [regex]::Replace($md, '(?<=\S) (\d{1,2})\. ([A-Z])', "`n" + '$1. $2')
```
This fires on any mid-sentence `" 2. Word"` pattern, including prose like *"See step 2. Verify the result."* or *"See section 3. Additional details."* The intent is to format numbered remediation steps onto separate lines, but the heuristic is too broad.

src/report/src/components/test-table/data-table.tsx
```diff
@@ -92,6 +92,13 @@ export function DataTable<TData extends Test, TValue>({
         return data;
     }, [data, pillar]);
 
+    // Default to High risk filter for Infrastructure pillar
+    React.useEffect(() => {
+        if (pillar === "Infrastructure" && pillarFilteredData.some(item => item.TestRisk === "High")) {
```
alexandair: **React `useEffect` for Infrastructure default risk filter has three problems**

```tsx
React.useEffect(() => {
    if (pillar === "Infrastructure" && pillarFilteredData.some(item => item.TestRisk === "High")) {
        setSelectedRisks(["High"]);
    }
}, [pillar, pillarFilteredData]);
```

1. **Unstable dependency:** `pillarFilteredData` is a `useMemo` value. Its reference changes when the memo recomputes, causing this effect to fire more often than intended. The dependency should be `pillar` alone (or a stable boolean derived from the memo).

2. **No reset on pillar switch:** When the user switches away from Infrastructure, the "High" filter remains active for the next pillar. There is no cleanup or reset path for `selectedRisks` in this effect.

3. **User-override overwritten:** If the user explicitly clears the High filter while viewing Infrastructure, any re-render that invalidates `pillarFilteredData` will re-apply it against their intent.

Fix: trigger the default only once per pillar switch, using a `useRef` initialized-flag or by resetting `selectedRisks` to `[]` whenever `pillar` changes, then setting the Infrastructure default in the same pass.

alexandair (changes requested):

src/powershell/private/core/ConvertTo-ZtMarkdown.ps1
praneeth-0000: This is fixed.

src/powershell/private/core/ConvertTo-ZtMarkdown.ps1
praneeth-0000: After analyzing ~50 descriptions & remediationSteps text from ARG query result, I've not found standalone sentences like "See step 2. Verify the result.". Even though tags are missing, the format always contained list of steps to follow. I feel we can keep the regex as is for now.

src/report/src/components/test-table/data-table.tsx
praneeth-0000: Updated the code to have dependency only on pillar.

praneeth-0000:
> Here is what I found after analyzing all 5 helper functions in `Test-Assessment.50001.ps1`:
> 
> `Get-NormalizedRisk`, `Get-DescriptionMarkdown`, `Get-ColumnFlags`, and `Get-McsbCells` are all correct. All bugs are in `Get-RowLinks` (lines 98–105).
> 
> 1. **Fix `SubMd` — blank link text when `subscriptionName` is empty** (_Bug 1_)
>    If the left-outer join on `resourcecontainers` misses (subscription not in the container list), `subscriptionName` is `''`. The current code produces `[](url)` — blank link text.
>    Fix: fall back to `subscriptionId` as display text when `subscriptionName` is empty.
> 2. **Fix `ResMd` — blank link text on per-row empty `resourceName`** (_Bug 2_)
>    `Get-ColumnFlags` sets `$showResource = $true` if _any_ row has a `resourceName`. Rows in the same recommendation that lack a name produce `[](https://portal.azure.com/#resource{id})` — blank, and a broken URL if `resourceId` is also empty.
>    Fix: fall back to `resourceId` (or a literal `'(resource)'`) as display text when `resourceName` is empty.
> 3. **Fix `PortalLinkMd` — MCSB-only rows may produce a relative portal URL** (_Bug 3_)
>    The Secure Score KQL (line ~218–220) applies an explicit `case()` to prefix any non-`https://` value before storing `azurePortalRecommendationLink`. The MCSB KQL (line ~307) does **not** — it uses `tostring(properties1.links.azurePortal)` verbatim. For MCSB-only rows, a link like `"portal.azure.com/#blade/..."` renders as a relative URL in the report and breaks navigation.
>    Fix: in `Get-RowLinks`, before building the `[View recommendation](...)` string, ensure the link starts with `https://`; prepend it otherwise.

This is resolved. Thanks.

src/powershell/private/core/ConvertTo-ZtMarkdown.ps1
praneeth-0000: Removed the file altogether, used rehype pkg to render html as is.

src/report/package.json
```diff
@@ -34,6 +34,8 @@
     "react-router-dom": "^6.30.3",
     "react-wrap-balancer": "^1.1.0",
     "recharts": "^2.12.7",
+    "rehype-raw": "^7.0.0",
+    "rehype-sanitize": "^6.0.0",
     "remark-gfm": "^4.0.0",
     "sonner": "^2.0.7",
```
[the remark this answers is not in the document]
praneeth-0000: resovled

src/report/src/components/test-table/data-table.tsx
```diff
@@ -500,14 +519,20 @@ export function DataTable<TData extends Test, TValue>({
...
                         </Card>
 
                         <Card>
                             <CardHeader><CardTitle>What was checked</CardTitle></CardHeader>
                             <CardContent>
-                                <Markdown className="prose max-w-fit dark:prose-invert" remarkPlugins={[remarkGfm]}>{selectedRow?.TestDescription}</Markdown>
+                                {selectedRow?.TestPillar === "Infrastructure"
+                                    ? <Markdown className="prose max-w-fit dark:prose-invert" remarkPlugins={[remarkGfm]} rehypePlugins={[rehypeRaw, rehypeSanitize]}>{selectedRow?.TestDescription}</Markdown>
+                                    : <Markdown className="prose max-w-fit dark:prose-invert" remarkPlugins={[remarkGfm]}>{selectedRow?.TestDescription}</Markdown>
+                                }
```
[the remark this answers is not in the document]
praneeth-0000: fixed

src/powershell/tests/Test-Assessment.50001.ps1
```diff
@@ -0,0 +1,720 @@
...
+    | extend assessmentDetails = parse_json(properties)
+    | extend resourceDetails = parse_json(assessmentDetails.resourceDetails)
+    | extend fullResourceType = tostring(resourceDetails.ResourceType)
+    | extend resourceType = tostring(split(fullResourceType, '/')[1])
+    | extend exportedTimestamp = now()
+    | extend recommendationId = id
+    | extend recommendationName = tostring(split(id, '/')[array_length(split(id, '/')) - 1])
+    | extend azurePortalRecommendationLink = case(
+        tostring(assessmentDetails.links.azurePortal) startswith "https://", tostring(assessmentDetails.links.azurePortal),
+        strcat("https://", tostring(assessmentDetails.links.azurePortal))
```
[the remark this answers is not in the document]
praneeth-0000: fixed

src/powershell/tests/Test-Assessment.50001.ps1
```diff
@@ -0,0 +1,725 @@
...
+    resourceSource =~ "aws" and isnotempty(tostring(properties1.resourceDetails.ConnectorId)), tostring(properties1.additionalData.ResourceType),
+    iff(regexResourceId[1] != "", regexResourceId[1], iff(regexResourceId[0] != "", regexResourceId[0], "subscriptions"))
+)
+| extend resourceName = tostring(regexResourceId[2])
+| extend recommendationName = name
+| extend recommendationDisplayName = tostring(iff(isnull(properties1.displayName) or isempty(properties1.displayName), properties.description, properties1.displayName))
+| extend description = tostring(properties1.metadata.description)
+| extend remediationSteps = tostring(properties1.metadata.remediationDescription)
+| extend severity = tostring(properties1.metadata.severity)
+| extend azurePortalRecommendationLink = tostring(properties1.links.azurePortal)
```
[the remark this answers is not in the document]
praneeth-0000: fixed

src/powershell/tests/Test-Assessment.50001.ps1
```diff
@@ -0,0 +1,728 @@
...
+    | where type == "microsoft.security/assessments"
+    | extend assessmentDetails = parse_json(properties)
+    | extend resourceDetails = parse_json(assessmentDetails.resourceDetails)
+    | extend fullResourceType = tostring(resourceDetails.ResourceType)
+    | extend resourceType = tostring(split(fullResourceType, '/')[1])
+    | extend exportedTimestamp = now()
+    | extend recommendationId = id
+    | extend recommendationName = tostring(split(id, '/')[array_length(split(id, '/')) - 1])
+    | extend azurePortalRecommendationLink = case(
+        tostring(assessmentDetails.links.azurePortal) startswith "https://", tostring(assessmentDetails.links.azurePortal),
```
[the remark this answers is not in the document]
praneeth-0000: fixed

src/powershell/private/tests/Invoke-ZtTests.ps1
alexandair, line 94: **Infrastructure tests silently produce 0 results without `-Preview`**

```powershell
$stablePillars = @('Identity', 'Devices', 'Network', 'Data')  # Infrastructure absent
$testsToRun = $testsToRun.Where{ $_.Pillar -in $stablePillars }
```
`Infrastructure` was correctly added to the `-Pillar` ValidateSet on line 56, but never added to `$stablePillars`. Result: `Invoke-ZtAssessment -Pillar Infrastructure` runs silently with 0 tests, no error, no warning. Copilot flagged this — author responded "Not needed." The silently-empty behavior is the real bug regardless of whether the pillar is preview-only. 

Fix: either add `Infrastructure` to `$stablePillars` if it's ready, or emit a warning when an explicitly-requested pillar yields 0 tests when `-Preview` is skipped.

### Verification Steps

1. Run `Invoke-ZtAssessment -Pillar Infrastructure` without `-Preview` — should emit a warning or error, not silently produce 0 results

src/powershell/tests/Test-Assessment.50001.ps1
```diff
@@ -0,0 +1,729 @@
...
+        $subText    = if (-not [string]::IsNullOrWhiteSpace($Row.subscriptionName)) { Get-SafeMarkdown $Row.subscriptionName } else { $Row.subscriptionId }
+        $resText    = if (-not [string]::IsNullOrWhiteSpace($Row.resourceName)) { Get-SafeMarkdown $Row.resourceName } else { $Row.resourceId }
+        $portalLink = $Row.azurePortalRecommendationLink
+        if (-not [string]::IsNullOrWhiteSpace($portalLink) -and -not $portalLink.StartsWith('https://')) {
+            $portalLink = "https://$portalLink"
+        }
+        [pscustomobject]@{
+            SubMd        = "[$subText]($subLink)"
+            ResMd        = "[$resText](https://portal.azure.com/#resource$($Row.resourceId))"
+            PortalLinkMd = if (-not [string]::IsNullOrWhiteSpace($portalLink)) { "[View recommendation]($portalLink)" } else { '' }
```
alexandair, line 93: **`Get-RowLinks` double-prefixes `http://` portal links → `https://http://...`**

```powershell
if (-not [string]::IsNullOrWhiteSpace($portalLink) -and -not $portalLink.StartsWith('https://')) {
    $portalLink = "https://$portalLink"
}
```
Both KQL queries now correctly preserve `http://` links as-is (the `case()` statements include `azurePortalLink startswith "http://", azurePortalLink`). The PowerShell guard only checks `.StartsWith('https://')`, so an `http://` link passes the condition and gets prefixed → `https://http://...`. 

Fix: change to `-not $portalLink.StartsWith('http')`.

### Verification Steps

1. Manually test `Get-RowLinks` with a portal link starting with `http://` — confirm no double-prefix

alexandair (changes requested):

src/powershell/private/core/Add-ZtTestResultDetail.ps1
```diff
@@ -173,19 +181,21 @@ function Add-ZtTestResultDetail {
...
-	if ($testMeta.Title) {
+	# For dynamic tests (where explicit params are passed), don't let testMeta override them.
+	# This allows functions like MDC that emit multiple rows with different titles/risk to work correctly.
+	if ($testMeta.Title -and -not $PSBoundParameters.ContainsKey('Title')) {
 		$docsTitle = $testMeta.Title
 	}
-	if ($testMeta.Category) {
+	if ($testMeta.Category -and -not $PSBoundParameters.ContainsKey('Category')) {
 		$category = $testMeta.Category
 	}
```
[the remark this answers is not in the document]
praneeth-0000: resolved

src/powershell/tests/Test-Assessment.50001.ps1
praneeth-0000: resolved

src/powershell/public/Invoke-ZtAssessment.ps1
```diff
@@ -259,21 +259,21 @@ $titleLine
...
+		Write-Host "The '$Pillar' pillar is currently in preview and requires the " -NoNewline -ForegroundColor Red
+		Write-Host "-Preview" -NoNewline -ForegroundColor Yellow
+		Write-Host " switch." -ForegroundColor Red
+		Write-Host
+		Write-Host "Please run the command again with the " -NoNewline -ForegroundColor White
+		Write-Host "-Preview" -NoNewline -ForegroundColor Yellow
+		Write-Host " parameter to assess the $Pillar pillar." -ForegroundColor White
+		Write-Host
+		return
+	}
```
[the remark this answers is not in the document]
praneeth-0000: SecOps and AI would be added seperatly.

src/report/src/components/test-table/data-table.tsx
```diff
@@ -210,6 +217,9 @@ export function DataTable<TData extends Test, TValue>({
 
     const [sheetOpen, setSheetOpen] = React.useState(false);
     const [selectedRow, setSelectedRow] = React.useState<Test | null>(null);
+    const mdRehypePlugins = selectedRow?.TestPillar === "Infrastructure"
+        ? [rehypeRaw, rehypeSanitize]
+        : [];
```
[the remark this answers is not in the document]
praneeth-0000: The data comes from Azure and default Schema strips out most of the html elements like <script>, onclick etc..

src/powershell/private/tests/Invoke-ZtTests.ps1
praneeth-0000: This is resolved.

src/powershell/public/Invoke-ZtAssessment.ps1
```diff
@@ -259,21 +259,21 @@ $titleLine
...
-	# 	Write-Host " switch." -ForegroundColor Red
-	# 	Write-Host
-	# 	Write-Host "Please run the command again with the " -NoNewline -ForegroundColor White
-	# 	Write-Host "-Preview" -NoNewline -ForegroundColor Yellow
-	# 	Write-Host " parameter to assess the $Pillar pillar." -ForegroundColor White
-	# 	Write-Host
-	# 	return
-	# }
+	$previewPillars = @('Infrastructure')
+	if ($Pillar -in $previewPillars -and -not $Preview) {
```
alexandair, line 264: **Preview gate bypassed when `-Pillar All` is used**

```powershell
$previewPillars = @('Infrastructure')
if ($Pillar -in $previewPillars -and -not $Preview) { ... return }
```

When a user runs `Invoke-ZtAssessment -Pillar All` (the default), `$Pillar` is `'All'`, so `'All' -in @('Infrastructure')` is `$false` and the gate never fires. Infrastructure tests run without any `-Preview` requirement. This defeats the purpose of the preview gate entirely — users get Infrastructure results whether they opt-in or not.

**Fix:** Also check if `$Pillar -eq 'All'` and warn the user.

alexandair (changes requested):

src/powershell/public/Invoke-ZtAssessment.ps1
praneeth-0000: Hi @alexandair , when -Pillar All is used, user will not get any warnings. If -Preview flag is no set, only stable pillars like 'Identity', 'Devices', 'Network', 'Data' get executed.

praneeth-0000: The code block in Invoke-ZtAssessment is only for UX and to prevent user from running preview pillars without -Preview parameter.

alexandair (approved):
LGTM

PR #1199, opened 2026-05-05
Bugfix 1151 - RoleAssignment import fails with timestamp parse error

ashwinikarke:
Added `ISO 8601 sentinel` values `("1900-01-01T00:00:00Z")` for all four timestamp fields to `RoleAssignment-model.json` so DuckDB always infers the correct format.

PR #1200, opened 2026-05-06
Bugfix - role-model-timestamps for RoleAssignmentScheduleInstance-model.json and RoleEligibilityScheduleInstance-model.json

ashwinikarke:


alexandair:
@ashwinikarke 
PR #1199 added to principal: `createdDateTime`, `refreshTokensValidFromDateTime`, `signInSessionsValidFromDateTime`, `assignedPlans`. PR #1200 adds those same four plus `renewedDateTime`. `renewedDateTime` is a `microsoft.graph.group` property (not a user property) — it appears in Graph responses when a group is the principal in a role assignment. The inconsistency is: if group principals can trigger this failure mode, then `RoleAssignment-model.json` (PR #1199) also needs `renewedDateTime`, but didn't get it.
This means that PR #1199 is now incomplete and needs an update.

alexandair (approved):
LGTM

SagarSathe (approved):

PR #1208, opened 2026-05-08
Bugfix 21816 - PIM privileged role assignment check should account for JIT access into role-assignable groups

### LINKED ISSUE #1177: PIM privileged role assignment check should account for JIT access into role-assignable groups
Check 21816 presumes the JIT boundary is at the role assignment layer. We do not account for a customer enforcing JIT elevation into a group, which in turn has a permanent assignment to a role.

Proposal: when a group is found with a permanent role assignment, query 
PIM eligibility to verify that group membership is managed via PIM for Groups.

### LINKED ISSUE #1293: Add PrivilegedAssignmentSchedule.Read.AzureADGroup and PrivilegedEligibilitySchedule.Read.AzureADGroup permissions to documentation
The documentation is currently missing two Microsoft Graph delegated permissions required by test 21816.

As part of the fix for #1177 (_PIM privileged role assignment check should account for JIT access into role-assignable groups_), test #1208 (21816) was updated to query two new PIM for Groups endpoints when a group with a permanent role assignment is found:

- `identityGovernance/privilegedAccess/group/eligibilityScheduleInstances` — checks whether group membership is managed via PIM eligibility schedules, confirming JIT elevation is enforced at the group layer.
- `identityGovernance/privilegedAccess/group/assignmentScheduleInstances` — checks for any standing active member assignments alongside eligibility schedules, to detect groups that mix PIM-eligible and permanently assigned members.

Calling these endpoints requires the following delegated scopes, which were added to `Get-ZtGraphScope`:

- `PrivilegedAssignmentSchedule.Read.AzureADGroup`
- `PrivilegedEligibilitySchedule.Read.AzureADGroup`

Could you please add these permissions to both pages:

* Public get-started page: https://learn.microsoft.com/en-us/security/zero-trust/assessment/get-started
* Repo docs page: https://github.com/microsoft/zerotrustassessment/blob/main/src/powershell/doc/readme.md

ashwinikarke:
## Problem

Check 21816 has two false negative cases:

1. **Groups with PIM for Groups configured were never verified further** — the original Q5 logic added non-GA groups directly to `$nonPIMPrivilegedGroups` without any PIM for Groups check at all. The original Q4 logic for GA groups expanded group members into `$permanentGAUserList` without checking whether JIT was enforced. Neither query verified that group membership was managed via PIM for Groups.

2. **Time-limited standing role assignments were not flagged** — the `$isPermanent` check used `assignmentType -eq 'Assigned' and $null -eq $pimAssignment.endDateTime`, missing any standing assignment that had an end date set. Entra PIM always sets an end date on active assignments made via the portal, so these were silently skipped.

## Changes

**`src/powershell/tests/Test-Assessment.21816.ps1`**

- **Q5 (non-GA groups) — PIM for Groups check added**: Groups are now checked via `eligibilityScheduleInstances` + `assignmentScheduleInstances`. A group is excluded only if eligibility exists **and** no `assignmentType eq 'assigned'` standing members are present
- **Q4 (GA groups) — PIM for Groups check added**: Same two-step logic. If the group enforces JIT with no standing members, member expansion is skipped. SPs found inside non-PIM-managed GA groups are surfaced in the non-PIM report table
- **Permanent assignment detection fixed**: Replaced direct array property access with `$hasPermAssignment = @($pimAssignment | Where-Object { $_.assignmentType -eq 'Assigned' -and $null -eq $_.endDateTime })` — correctly handles multiple schedule instances and restores the `endDateTime` null check so time-limited active PIM assignments are not flagged
- **`assignmentType` display fixed**: Uses `($hasPermAssignment | Select-Object -First 1).assignmentType` — guaranteed to be an `Assigned` record
- **Service principal guard added**: Both non-GA and GA loops now have explicit `elseif` for groups (PIM for Groups check) and `else` for SPs/other types (always flagged, rendered with `ManagedAppMenuBlade` link)
- **`-Select` restored with `appId` added**: Both `directoryRoles/{id}/members` calls use `-Select 'id,displayName,userPrincipalName,appId'`; GA group expansion `groups/{id}/members` `-Select` removed to return `appId` and `@odata.type` for SP detection
- **`$permanentGAGroupList` removed**: Was added and immediately removed in both branches — dead state
- **Report**: SPs rendered with `ManagedAppMenuBlade` link (`objectId` + `appId`); groups keep `GroupDetailsMenuBlade` link
- **User-facing messages** updated to match spec wording exactly

**`src/powershell/public/Get-ZtGraphScope.ps1`**

Two new scopes added — both required together:

- `PrivilegedEligibilitySchedule.Read.AzureADGroup` — required to call `eligibilityScheduleInstances` to verify JIT is configured on the group
- `PrivilegedAssignmentSchedule.Read.AzureADGroup` — required to call `assignmentScheduleInstances` to detect standing active members who have bypassed JIT


src/powershell/tests/Test-Assessment.21816.ps1
```diff
@@ -71,16 +71,18 @@ function Test-Assessment-21816 {
...
             foreach ($member in $roleMembers) {
                 # Check if assignment is managed by PIM
                 $pimAssignment = Invoke-ZtGraphRequest -RelativeUri 'roleManagement/directory/roleAssignmentScheduleInstances' -Filter "principalId eq '$($member.id)' and roleDefinitionId eq '$($role.templateId)'" -ApiVersion beta
                 Write-PSFMessage "PIM assignment check for $($member.displayName): Found=$($pimAssignment.Count) results" -Level Verbose
 
-                if (-not $pimAssignment -or ($pimAssignment.assignmentType -eq 'Assigned' -and $null -eq $pimAssignment.endDateTime)) {
-                    # Not managed by PIM or permanent assignment
+                # Array member access — truthy when any Assigned record exists; Activated (JIT self-activation) records are excluded
+                if (-not $pimAssignment -or $pimAssignment.assignmentType -eq 'Assigned') {
+                    # Not managed by PIM or standing active assignment
```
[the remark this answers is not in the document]
ashwinikarke: Not valid. PowerShell's -eq on an array acts as a filter, not a boolean comparison — it returns matching elements. @('Activated','Activated') -eq 'Assigned' returns @() (falsy), not @($false,$false). The described false-positive cannot occur.

src/powershell/public/Get-ZtGraphScope.ps1
```diff
@@ -37,6 +37,8 @@
         'Policy.Read.ConditionalAccess'
         'Policy.Read.PermissionGrant'
         'PrivilegedAccess.Read.AzureAD'
+        'PrivilegedAssignmentSchedule.Read.AzureADGroup'
+        'PrivilegedEligibilitySchedule.Read.AzureADGroup'
         'Reports.Read.All'
```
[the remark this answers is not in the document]
ashwinikarke: Ignore

src/powershell/tests/Test-Assessment.21816.ps1
```diff
@@ -107,14 +125,16 @@ function Test-Assessment-21816 {
...
 
         foreach ($member in $gaMembers) {
             # Check if GA assignment is managed by PIM
             $pimAssignment = Invoke-ZtGraphRequest -RelativeUri 'roleManagement/directory/roleAssignmentScheduleInstances' -Filter "principalId eq '$($member.id)' and roleDefinitionId eq '$globalAdminRoleId'" -ApiVersion beta
 
-            if (-not $pimAssignment -or ($pimAssignment.assignmentType -eq 'Assigned' -and $null -eq $pimAssignment.endDateTime)) {
-                # Permanent GA assignment found
+            # Array member access — truthy when any Assigned record exists; Activated (JIT self-activation) records are excluded
+            if (-not $pimAssignment -or $pimAssignment.assignmentType -eq 'Assigned') {
+                # Standing active GA assignment found
```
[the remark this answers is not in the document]
ashwinikarke: Not valid — same misunderstanding as the previous comment. PowerShell's -eq applied to an array on the left is a filter, not a per-element boolean comparison. It returns matching elements, not $true/$false values.

src/powershell/tests/Test-Assessment.21816.ps1
```diff
@@ -123,30 +143,49 @@ function Test-Assessment-21816 {
...
+                    } catch {
+                        Write-PSFMessage "Could not check PIM for Groups for group $($member.displayName): $($_.Exception.Message). Treating as non-PIM managed." -Level Warning
+                    }
+                    if ($pimForGroupsSchedules -and $pimForGroupsPermanentMembers.Count -eq 0) {
+                        # Group enforces JIT via PIM for Groups with no permanent members - exclude from PermanentGAGroupList
+                        $permanentGAGroupList = $permanentGAGroupList | Where-Object { $_.id -ne $member.id }
+                        Write-PSFMessage "Group $($member.displayName) uses PIM for Groups with no permanent members, excluding from permanent GA list" -Level Verbose
+                    } else {
+                        # Group does not use PIM for Groups - get members and add to permanentGAUserList (Q5)
+                        $groupMembers = Invoke-ZtGraphRequest -RelativeUri "groups/$($member.id)/members" -Select 'userPrincipalName,displayName,id,onPremisesSyncEnabled' -ApiVersion beta
```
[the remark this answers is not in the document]
ashwinikarke: fixed

src/powershell/tests/Test-Assessment.21816.ps1
```diff
@@ -123,30 +143,49 @@ function Test-Assessment-21816 {
...
+                        }
+                    } catch {
+                        Write-PSFMessage "Could not check PIM for Groups for group $($member.displayName): $($_.Exception.Message). Treating as non-PIM managed." -Level Warning
+                    }
+                    if ($pimForGroupsSchedules -and $pimForGroupsPermanentMembers.Count -eq 0) {
+                        # Group enforces JIT via PIM for Groups with no permanent members - exclude from PermanentGAGroupList
+                        $permanentGAGroupList = $permanentGAGroupList | Where-Object { $_.id -ne $member.id }
+                        Write-PSFMessage "Group $($member.displayName) uses PIM for Groups with no permanent members, excluding from permanent GA list" -Level Verbose
+                    } else {
+                        # Group does not use PIM for Groups - get members and add to permanentGAUserList (Q5)
```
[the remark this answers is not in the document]
ashwinikarke: fixed

src/powershell/tests/Test-Assessment.21816.ps1
```diff
@@ -71,16 +71,18 @@ function Test-Assessment-21816 {
...
 
             foreach ($member in $roleMembers) {
                 # Check if assignment is managed by PIM
                 $pimAssignment = Invoke-ZtGraphRequest -RelativeUri 'roleManagement/directory/roleAssignmentScheduleInstances' -Filter "principalId eq '$($member.id)' and roleDefinitionId eq '$($role.templateId)'" -ApiVersion beta
                 Write-PSFMessage "PIM assignment check for $($member.displayName): Found=$($pimAssignment.Count) results" -Level Verbose
 
-                if (-not $pimAssignment -or ($pimAssignment.assignmentType -eq 'Assigned' -and $null -eq $pimAssignment.endDateTime)) {
-                    # Not managed by PIM or permanent assignment
+                # Array member access — truthy when any Assigned record exists; Activated (JIT self-activation) records are excluded
+                if (-not $pimAssignment -or $pimAssignment.assignmentType -eq 'Assigned') {
```
alexandair: **Spec deviation: `endDateTime` check removed (false positives)**

**Location:** Line 84 (non-GA loop) and ~line 145 (GA loop)

**Old condition:**
```powershell
if (-not $pimAssignment -or ($pimAssignment.assignmentType -eq 'Assigned' -and $null -eq $pimAssignment.endDateTime)) {
```

**New condition:**
```powershell
if (-not $pimAssignment -or $pimAssignment.assignmentType -eq 'Assigned') {
```

The spec (Check Query section) is explicit:

> Mark the check as Failed if … both of the following values are set:
> - `AssignmentType == Assigned`
> - `EndDateTime == null`

A **time-limited active PIM assignment** (`assignmentType == 'Assigned'` with a non-null `endDateTime`, e.g. a 30-day scoped active assignment) will now be incorrectly flagged as "not managed by PIM". This is a real scenario — PIM admins can create active assignments with expiration dates.

Additionally, the old code had its own array-handling bug (member enumeration made the `-and` between two arrays unreliable). The correct fix should use per-item `Where-Object`:

```powershell
$hasPermAssignment = @($pimAssignment | Where-Object { $_.assignmentType -eq 'Assigned' -and $null -eq $_.endDateTime })
if (-not $pimAssignment -or $hasPermAssignment.Count -gt 0) {
```

This needs to be applied in both the non-GA loop and the GA loop.

src/powershell/tests/Test-Assessment.21816.ps1
```diff
@@ -89,13 +91,29 @@ function Test-Assessment-21816 {
...
-                        $nonPIMPrivilegedGroups += $memberInfo
+                        # Q5: Check if the group uses PIM for Groups to enforce JIT member access
+                        $pimForGroupsSchedules = $null
+                        $pimForGroupsPermanentMembers = @()
+                        try {
+                            $pimForGroupsSchedules = Invoke-ZtGraphRequest -RelativeUri 'identityGovernance/privilegedAccess/group/eligibilityScheduleInstances' -Filter "groupId eq '$($member.id)' and accessId eq 'member'" -ApiVersion beta -ErrorAction Stop
+                            if ($pimForGroupsSchedules) {
+                                # Also check for permanent active member assignments alongside eligibility schedules
+                                $allGroupAssignments = Invoke-ZtGraphRequest -RelativeUri 'identityGovernance/privilegedAccess/group/assignmentScheduleInstances' -Filter "groupId eq '$($member.id)' and accessId eq 'member'" -ApiVersion beta -ErrorAction Stop
+                                $pimForGroupsPermanentMembers = @($allGroupAssignments | Where-Object { $_.assignmentType -eq 'assigned' })
```
alexandair, line 114: **Same `endDateTime` gap in PIM-for-Groups permanent member detection**

**Location:** Inside both `try` blocks:
```powershell
$pimForGroupsPermanentMembers = @($allGroupAssignments | Where-Object { $_.assignmentType -eq 'assigned' })
```

A time-limited active group member (`assignmentType == 'assigned'`, `endDateTime != null`) is treated the same as a permanent member. This causes the group to be retained in the failure list even though the member's access will expire. The MS Docs confirm `endDateTime` is available on `privilegedAccessGroupAssignmentScheduleInstance`. The fix:

```powershell
$pimForGroupsPermanentMembers = @($allGroupAssignments | Where-Object { $_.assignmentType -eq 'assigned' -and $null -eq $_.endDateTime })
```

src/powershell/tests/Test-Assessment.21816.ps1
```diff
@@ -89,13 +91,29 @@ function Test-Assessment-21816 {
                         roleDefinitionId = $role.id
                         roleName = $role.displayName
                         isPrivileged = $true
-                        assignmentType = if ($pimAssignment) { $pimAssignment.assignmentType } else { 'Not in PIM' }
+                        assignmentType = if ($pimAssignment) { ($pimAssignment | Select-Object -First 1).assignmentType } else { 'Not in PIM' }
                     }
 
                     if ($member.'@odata.type' -eq '#microsoft.graph.user') {
```
alexandair, line 101: **Service principals silently treated as groups**

The `else` branch in the non-GA loop handles any non-user principal as a group:

```powershell
if ($member.'@odata.type' -eq '#microsoft.graph.user') {
    $nonPIMPrivilegedUsers += $memberInfo
} else {
    # Q5: Check if the group uses PIM for Groups...
    $pimForGroupsSchedules = Invoke-ZtGraphRequest -RelativeUri 'identityGovernance/privilegedAccess/group/eligibilityScheduleInstances'...
```

A service principal in a privileged role will flow into this branch. The PIM-for-Groups query will return no schedules for an SP (`$pimForGroupsSchedules = $null`), so `-not $pimForGroupsSchedules` is true, and the SP lands in `$nonPIMPrivilegedGroups`. It will then be displayed in the report with a **group portal link** (`/GroupDetailsMenuBlade/...`), which is incorrect for an SP. 

Suggest adding an explicit SP guard:
```powershell
if ($member.'@odata.type' -eq '#microsoft.graph.user') {
    $nonPIMPrivilegedUsers += $memberInfo
} elseif ($member.'@odata.type' -eq '#microsoft.graph.group') {
    # PIM-for-Groups check...
} else {
    # Service principal or other type - always non-PIM
    $nonPIMPrivilegedGroups += $memberInfo
}
```

src/powershell/tests/Test-Assessment.21816.ps1
```diff
@@ -71,16 +71,18 @@ function Test-Assessment-21816 {
         if ($directoryRole) {
             Write-PSFMessage "Found directory role instance for $($role.displayName)" -Level Verbose
             # Get members of this role
-            $roleMembers = Invoke-ZtGraphRequest -RelativeUri "directoryRoles/$($directoryRole.id)/members" -Select 'userPrincipalName,displayName,id' -ApiVersion beta
+            # Raw call required — Get-ZtRoleMember flattens group members; group objects are needed to check PIM for Groups
```
alexandair, line 73: **Removing `-Select` returns unnecessary payload**

```powershell
# Before
$roleMembers = Invoke-ZtGraphRequest -RelativeUri "directoryRoles/$($directoryRole.id)/members" -Select 'userPrincipalName,displayName,id' -ApiVersion beta
# After
$roleMembers = Invoke-ZtGraphRequest -RelativeUri "directoryRoles/$($directoryRole.id)/members" -ApiVersion beta
```

The justification in the comment is that `@odata.type` is needed to identify groups. However, `@odata.type` is an OData annotation and is included in polymorphic responses **even when `$select` is used**. Removing `-Select` returns all properties on every member, which increases response payload significantly for large roles. Consider restoring `-Select` and adding `@odata.type` explicitly, or just testing that the annotation is preserved with a select.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.21816.ps1
ashwinikarke: fixed

src/powershell/tests/Test-Assessment.21816.ps1
ashwinikarke: The `group/assignmentScheduleInstances` API marks `endDateTime` as a **Required** field per MS Docs, unlike `roleAssignmentScheduleInstances` where it can be null for permanent assignments. In live testing, `endDateTime` is always non-null in this API regardless of whether the membership is permanent or time-limited. Adding `$null -eq $_.endDateTime` would cause permanent members to never be detected.

The correct distinction in this API is `assignmentType`, not `endDateTime`:
- `assignmentType -eq 'assigned'` = admin-placed standing active membership (permanent access, always flagged)
- `assignmentType -eq 'activated'` = JIT self-activation of an eligibility (temporary, not standing access — correctly excluded)

`assignmentType -eq 'assigned'` alone is therefore the correct signal for standing active membership in this API.

src/powershell/tests/Test-Assessment.21816.ps1
ashwinikarke: fixed

src/powershell/tests/Test-Assessment.21816.ps1
ashwinikarke: fixed

src/powershell/tests/Test-Assessment.21816.ps1
alexandair: **Summary table label is now misleading**

The summary table (line ~265) reads:
```powershell
$mdInfo += "| Non-PIM privileged groups | $($nonPIMPrivilegedGroups.Count) |`n"
```
`$nonPIMPrivilegedGroups` now contains both groups **and** service principals (direct SP role members + SPs found inside GA groups). The label should be updated, e.g. `Non-PIM privileged groups / service principals`.

src/powershell/tests/Test-Assessment.21816.ps1
```diff
@@ -107,48 +134,92 @@ function Test-Assessment-21816 {
...
+                        }
+                    } catch {
+                        Write-PSFMessage "Could not check PIM for Groups for group $($member.displayName): $($_.Exception.Message). Treating as non-PIM managed." -Level Warning
+                    }
+                    if ($pimForGroupsSchedules -and $pimForGroupsPermanentMembers.Count -eq 0) {
+                        # Group enforces JIT via PIM for Groups with no permanent members - skip member expansion
+                        Write-PSFMessage "Group $($member.displayName) uses PIM for Groups with no permanent members, excluding from permanent GA list" -Level Verbose
+                    } else {
+                        # Group does not use PIM for Groups - get members and add to permanentGAUserList (Q4)
+                        $groupMembers = Invoke-ZtGraphRequest -RelativeUri "groups/$($member.id)/members" -ApiVersion beta
```
alexandair: **No `-Select` on `groups/{id}/members` in the GA group expansion path**

```powershell
$groupMembers = Invoke-ZtGraphRequest -RelativeUri "groups/$($member.id)/members" -ApiVersion beta
```
This returns all properties for every member of the group. The corresponding call for eligible GA groups correctly uses `-Select`. Suggest adding `-Select 'id,displayName,userPrincipalName,onPremisesSyncEnabled,appId'` here for consistency and to reduce payload.

alexandair (changes requested):

alexandair (approved):
LGTM

PR #1211, opened 2026-05-11
Bugfix/issue 1194: fix issues related to test 21835

### LINKED ISSUE #1194: Test-21835: performance, correctness, and reporting cleanups for emergency access account check required
issues:
- Duplicate authentication methods rendered in Emergency access accounts table
- O(n²) role lookups in CA policy loop
- No short-circuit when there are zero enabled CA policies
- ~~memberOf is non-transitive for directory roles
users/{id}/memberOf/microsoft.graph.directoryRole only returns directly assigned roles. Users effectively assigned roles via role-assignable groups are missed, making CA includeRoles/excludeRoles evaluation (and the new "CA policies missing exclusion" column) inaccurate. Switch to transitiveMemberOf/microsoft.graph.directoryRole~~ --> to be tracked separately in issue #1250 
- Sort-Object -Property CAExcluded, PhishingResistant, CloudOnly sorts on emoji glyphs.
Pre-existing, not introduced here, but the comment "users that have passed every criteria first" is true only because U+2705 (✅) < U+274C (❌) in code-point order and PowerShell's culture-sensitive default happens to agree on Windows. Brittle. Consider sorting by booleans stored alongside the row, but out of scope for this PR.
- Status thresholds disagree with the spec.
$accountCount -ge 2 -and $accountCount -le 4 → Pass, > 4 → Fail. Spec says == 2 → Pass, > 2 → Investigate, < 2 → Fail, and the test currently has no Investigate path. Pre-existing — out of scope for this PR but worth a follow-up.
Some of these issues were reported in PR review comments: https://github.com/microsoft/zerotrustassessment/pull/1182

sandeepjha000:
- Align pass/fail thresholds with spec: <2 Fail, ==2 Pass, >2 Investigate (via -CustomStatus Investigate, matching Test-Assessment-21816).
- Update Pass/Fail messages to spec wording; Investigate message focuses on break-glass scope.
- Use splatting for Add-ZtTestResultDetail to conditionally include CustomStatus.
- Precompute $userRoleTemplateIds to drop the O(n·m) Where-Object role lookup per CA policy.
- Short-circuit per-user CA evaluation when there are zero enabled CA policies.
- ~~Use transitiveMemberOf/microsoft.graph.directoryRole so role-assignable group memberships are included.~~ --> to be taken care in separate issue #1250 
- Sort the permanent GA table on underlying booleans instead of emoji glyphs.
- Deduplicate authentication methods in the Emergency access accounts table.

Fixes #1194

### DIFF
--- a/src/powershell/tests/Test-Assessment.21835.ps1
+++ b/src/powershell/tests/Test-Assessment.21835.ps1
@@ -141,6 +141,20 @@ WHERE vr.roleDefinitionId = '62e90394-69f5-4237-9190-012177145e10'
     # Store CA policy info for ALL permanent GAs (not just candidates)
     $gaCAInfo = @{}
 
+    if ($enabledCAPolicies.Count -eq 0) {
+        # No enabled CA policies in the tenant: there is nothing to evaluate per user, and
+        # in particular nothing to exclude from. Vacuously, every user is "excluded from all"
+        # enabled policies. Skip the per-user Graph calls entirely.
+        Write-PSFMessage 'No enabled CA policies found; skipping per-user CA evaluation.' -Level Verbose
+        foreach ($user in $permanentGAUsers) {
+            $gaCAInfo[$user.id] = @{
+                PoliciesTargeting        = 0
+                ExcludedFromAll          = $true
+                PoliciesMissingExclusion = [System.Collections.Generic.List[object]]::new()
+            }
+        }
+    }
+    else {
     foreach ($user in $permanentGAUsers) {
         Write-PSFMessage "Checking CA policy targeting for: $($user.userPrincipalName)" -Level Verbose
 
@@ -244,6 +258,7 @@ WHERE vr.roleDefinitionId = '62e90394-69f5-4237-9190-012177145e10'
             ExcludedFromAll          = $excludedFromAll
             PoliciesMissingExclusion = $policiesMissingExclusion
         }
+    }
     }
 
     # Determine emergency access accounts: candidates that are excluded from all enabled CA policies
--- a/src/powershell/tests/Test-Assessment.21835.ps1
+++ b/src/powershell/tests/Test-Assessment.21835.ps1
@@ -304,7 +304,7 @@ WHERE vr.roleDefinitionId = '62e90394-69f5-4237-9190-012177145e10'
             $syncStatus = if ($account.onPremisesSyncEnabled -ne $true) { 'No' } else { 'Yes' }
             $authMethodDisplay = ($account.AuthenticationMethods | ForEach-Object {
                 $_ -replace '#microsoft.graph.', '' -replace 'AuthenticationMethod', ''
-            }) -join ', '
+            } | Select-Object -Unique) -join ', '
 
             $portalLink = "https://entra.microsoft.com/#view/Microsoft_AAD_UsersAndTenants/UserProfileMenuBlade/~/overview/userId/$($account.Id)"
 
--- a/src/powershell/tests/Test-Assessment.21835.ps1
+++ b/src/powershell/tests/Test-Assessment.21835.ps1
@@ -328,11 +328,13 @@ WHERE vr.roleDefinitionId = '62e90394-69f5-4237-9190-012177145e10'
             $cloudOnlyEmoji = if ($isCloudOnly) { '✅' } else { '❌' }
 
             # Check if excluded from all enabled CA policies (using per-user CA info, not emergency account membership)
-            $caExcludedEmoji = if ($gaCAInfo.ContainsKey($user.id) -and $gaCAInfo[$user.id].ExcludedFromAll) { '✅' } else { '❌' }
+            $isCAExcluded = ($gaCAInfo.ContainsKey($user.id) -and $gaCAInfo[$user.id].ExcludedFromAll)
+            $caExcludedEmoji = if ($isCAExcluded) { '✅' } else { '❌' }
 
             # Check if has phishing-resistant auth only
             $candidate = $emergencyAccountCandidates | Where-Object { $_.Id -eq $user.id }
-            $phishingResistantEmoji = if ($candidate) { '✅' } else { '❌' }
+            $isPhishingResistant = [bool]$candidate
+            $phishingResistantEmoji = if ($isPhishingResistant) { '✅' } else { '❌' }
 
             # Build CA policies missing exclusion cell
             if (-not $gaCAInfo.ContainsKey($user.id)) {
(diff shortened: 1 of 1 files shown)


src/powershell/tests/Test-Assessment.21835.ps1
alexandair: `$Database` parameter is missing.

alexandair, line 48: ### Root cause

`vwRole` is built by Export-Database.ps1#L102-L118 as a `UNION ALL` over:

1. Direct role assignments from `RoleAssignment` / `RoleAssignmentScheduleInstance`, and
2. Flattened members from the matching `*Group` tables (role-assignable groups assigned the role).

`UNION ALL` does not dedupe. A user who holds the GA role both directly **and** via membership of a role-assignable group (or via multiple role-assignable groups) appears as N rows.

The 21835 query at Test-Assessment.21835.ps1#L45-L59 filters by `roleDefinitionId`, `privilegeType = 'Permanent'`, and `@odata.type = '#microsoft.graph.user'`, but does not collapse on `principalId`. Duplicates flow through to `$permanentGAUsers`.

### Impact

Each duplicate row is treated as an independent user:

- `$emergencyAccountCandidates` gets pushed twice in L114-L122.
- `$emergencyAccessAccounts` is built by iterating candidates and accepting any whose `$gaCAInfo[$candidate.Id].ExcludedFromAll` is true (L272-L283) — `$gaCAInfo` is keyed by id (dedupes there), but the **candidate list is not deduped**, so both copies pass through.
- `$accountCount = $emergencyAccessAccounts.Count` is therefore inflated.

With the new exact-2 threshold this PR introduces, the failure modes become:

- 1 real break-glass account, duplicated once → `accountCount = 2` → **false Pass**.
- 2 real break-glass accounts, one duplicated → `accountCount = 3` → **false Investigate**.

The "All permanent Global Administrators" table also lists the user twice.

### Fix

Dedupe at the SQL layer — cheapest and benefits other downstream code (`Get-ZtUserAuthenticationMethod` and the transitive-group Graph calls run once per row today, so dedupe also saves API calls):

```sql
SELECT
    vr.principalId as id,
    ANY_VALUE(vr.principalDisplayName) as displayName,
    ANY_VALUE(vr.userPrincipalName)    as userPrincipalName,
    ANY_VALUE(vr.privilegeType)        as privilegeType,
    ANY_VALUE(u.onPremisesSyncEnabled) as onPremisesSyncEnabled,
    ANY_VALUE(vr."@odata.type")        as "@odata.type"
FROM vwRole vr
LEFT JOIN "User" u ON vr.principalId = u.id
WHERE vr.roleDefinitionId = '62e90394-69f5-4237-9190-012177145e10'
    AND vr.privilegeType = 'Permanent'
    AND vr."@odata.type" = '#microsoft.graph.user'
GROUP BY vr.principalId
```

(DuckDB supports `ANY_VALUE`; alternatively `SELECT DISTINCT` works because all selected non-id columns are deterministic per `principalId`.)

PowerShell-side fallback if SQL change is undesired:

```powershell
$permanentGAUsers = @(Invoke-DatabaseQuery -Database $Database -Sql $sql) |
    Group-Object id | ForEach-Object { $_.Group[0] }
```

Either way, dedupe **before** the auth-method and CA-evaluation loops so the savings carry through.

### Related sibling tests to flag (out of scope, but worth a follow-up issue)

The same `vwRole` duplication risk exists for any test counting rows: Test-Assessment.21813.ps1, Test-Assessment.21815.ps1, Test-Assessment.21836.ps1, Test-Assessment.22128.ps1, Test-Assessment.25383.ps1. Worth a quick audit — but don't expand this PR's scope.

alexandair: ### What the spec says

21835.md "Check Query" Step 2:
> User authentication methods returned by Q4 **only** has @odata.type property values of `microsoft.graph.fido2AuthenticationMethod` and/or `microsoft.graph.x509CertificateAuthenticationMethod`

21835.md "Test evaluation logic":
> exactly two accounts meet all criteria: cloud-only, **FIDO2/CBA auth methods only**, excluded from all CA policies

### What the code does

Test-Assessment.21835.ps1#L94-L110:

```powershell
$hasPhishingResistant = $false
foreach ($method in $authMethods) {
    $methodType = $method.'@odata.type'
    $authMethodTypes += $methodType
    if ($methodType -eq '#microsoft.graph.fido2AuthenticationMethod' -or
        $methodType -eq '#microsoft.graph.x509CertificateAuthenticationMethod') {
        $hasPhishingResistant = $true
    }
}
if ($hasPhishingResistant) { ...candidate... }
```

This is "has ≥1 FIDO2 or CBA method", not "only FIDO2/CBA". The `#579` comment correctly identifies that the original strict reading is broken (every account always has `passwordAuthenticationMethod`), but the fix overshot — it relaxed the predicate entirely instead of excluding `passwordAuthenticationMethod` from the set under test.

### Concrete false-positive

A normal Global Administrator with FIDO2 + Microsoft Authenticator + phone passes the current check and lands in `$emergencyAccountCandidates`. If they are also excluded from all CA policies (e.g. by sitting in a broadly-excluded admin group), they get counted as an emergency access account. Combined with the new exact-2 threshold this PR introduces, this is precisely the kind of input that flips Pass↔Investigate↔Fail on a single misclassified user.

### Recommended fix (code-side)

Treat the password method as inert and require everything else to be FIDO2 or CBA:

```powershell
$phishingResistantTypes = @(
    '#microsoft.graph.fido2AuthenticationMethod'
    '#microsoft.graph.x509CertificateAuthenticationMethod'
)
$ignorableTypes = @(
    '#microsoft.graph.passwordAuthenticationMethod'
)

$authMethodTypes = @($authMethods | ForEach-Object { $_.'@odata.type' })
$relevantTypes   = @($authMethodTypes | Where-Object { $_ -notin $ignorableTypes })

$hasPhishingResistant = $relevantTypes.Count -gt 0 -and
    -not ($relevantTypes | Where-Object { $_ -notin $phishingResistantTypes })
```

This restores the spec's "only FIDO2/CBA" intent while still working around the always-present password method that broke the original check.

### Recommended path

This is pre-existing, was deliberately introduced for #579, and is a behavior change with security implications. Two reasonable options:

1. **Tighten the code in this PR** (preferred, since the PR already touches behaviorally-adjacent code and the new exact-2 threshold amplifies the risk). Add a regression note referencing #579.
2. **File a separate issue** ("Test 21835 candidate predicate diverges from spec — accepts FIDO2 + non-phishing-resistant combinations") and explicitly leave it out of this PR with a comment. Acceptable but leaves the false-positive in `dev` while #1211 ships an exact threshold that makes it bite harder.

Either way, the spec and code need to agree — and the spec wording is the right behavior, so the code should be the one to move.

alexandair, line 18: Pre-existing metadata mismatch: 
Code declares `Category = 'Application management'` and `ImplementationCost = 'High'`, but the spec says `Privileged access` and `Medium`. This will misclassify the report metadata unless the spec is intentionally stale.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.21835.ps1
sandeepjha000: fixed in commit: [b95097c](https://github.com/microsoft/zerotrustassessment/pull/1211/commits/b95097ccae3367765955fbca3691c9647e2f8bc7)

sandeepjha000: fixed in commit: [b95097c](https://github.com/microsoft/zerotrustassessment/pull/1211/commits/b95097ccae3367765955fbca3691c9647e2f8bc7)

sandeepjha000: fixed in commit: [b95097c](https://github.com/microsoft/zerotrustassessment/pull/1211/commits/b95097ccae3367765955fbca3691c9647e2f8bc7)

sandeepjha000: fixed in commit: [b95097c](https://github.com/microsoft/zerotrustassessment/pull/1211/commits/b95097ccae3367765955fbca3691c9647e2f8bc7)

alexandair (approved):
LGTM

PR #1219, opened 2026-05-15
Add AI pillar support to DB export pipeline

### LINKED ISSUE #1218: Add AI pillar support to DB export pipeline
The AI pillar assessment checks require additional properties `(sponsors, agentIdentityBlueprintId)` to be included in the Graph API exports, and the `AI` pillar needs to be registered across the export configuration and DB loading pipeline.

Without these changes:

- The relevant navigation properties are not fetched from Graph when the AI pillar is selected
- No data is exported and no tables are loaded into the DB when -Pillar AI is passed

ashwinikarke:
The AI pillar assessment checks require additional properties to be included in the Graph API exports, and the AI pillar needs to be registered across the export configuration and DB loading pipeline so that `-Pillar AI` works end-to-end.

### Changes

**Export configuration (`export-tenant.config.psd1`)**
- `Application`: added `sponsors` to `RelatedPropertyNames`; added `AI` to `Pillar`
- `ServicePrincipal`: added `agentIdentityBlueprintId` to `$select`; added `sponsors` to `RelatedPropertyNames`; added `AI` to `Pillar`
- `SignIn`, `RoleDefinition`, `RoleAssignment`, `RoleAssignmentGroup`, `RoleEligibilityScheduleInstance`, `RoleEligibilityScheduleInstanceGroup`: added `AI` to `Pillar`

**DB loading (`Export-Database.ps1`)**
- Added an `AI` pillar block that loads the required tables into DuckDB when `-Pillar AI` is passed

### Tables that can be loaded for the AI pillar specs

| Spec | Tables |
|---|---|
| 61006 | `RoleDefinition`, `RoleAssignment`, `RoleAssignmentGroup`, `RoleEligibilityScheduleInstance`, `RoleEligibilityScheduleInstanceGroup` |
| 61008  | `ServicePrincipal` |
| 61011 | `Application`, `ServicePrincipal`, `SignIn` |
| 61013 | `Application`, `ServicePrincipal` |
| 61014 | `ServicePrincipal` |

### DIFF
--- a/src/powershell/assets/export-tenant.config.psd1
+++ b/src/powershell/assets/export-tenant.config.psd1
@@ -75,10 +75,10 @@ Note: Avoid using the same names as used for the "General Parameters" section of
 	Name = 'Application'
 	Uri = 'beta/applications'
 	QueryString = '$top=999'
-	RelatedPropertyNames = @()
+	RelatedPropertyNames = @('sponsors')
 	Type = 'Default' # PrivilegedGroup
 
-	Pillar = @('Identity', 'Network')
+	Pillar = @('Identity', 'Network', 'AI')
 	# Environment = $null # 'Global'
 	# IncludePlan = @('Free') # P2, Governance
 	# ExcludePlan = @('Free') # Free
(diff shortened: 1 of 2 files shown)


alexandair (changes requested):

**Issue: New-ViewRole called 3× when $Pillar = 'All'**

GitHub Copilot: The cleanest fix is to **pull `New-ViewRole` out of every pillar block** and call it once at the end, guarded by the combined pillar condition:

```powershell
# After all pillar blocks...

if ($Pillar -in ('All', 'Identity', 'Network', 'AI')) {
    New-ViewRole -Database $database
}

$database
```

And remove the three individual `New-ViewRole` calls from the Identity, Network, and AI blocks.

**Why this is safe:** `New-ViewRole` only depends on `RoleDefinition`, `RoleAssignmentScheduleInstance` / `RoleAssignment`, and `RoleEligibilityScheduleInstance` being present. Those tables are guaranteed to be loaded whenever any of `Identity`, Network, or AI was processed — the exact same conditions that gate the new single call.

**Single-pillar cases still work correctly:**
- `-Pillar Identity` → tables loaded by Identity block → `New-ViewRole` called once at the end ✅  
- `-Pillar Network` → tables loaded by Network block → view built once ✅  
- `-Pillar AI` → tables loaded by AI block → view built once ✅  
- `-Pillar All` → all three blocks load tables, but `New-ViewRole` runs exactly once ✅  
- `-Pillar Devices` → condition is false → `New-ViewRole` skipped (no role tables) ✅

This is a two-line change per pillar block (remove `New-ViewRole`, add one call at the end) and eliminates the redundant `Get-ZtLicenseInformation` calls and triple view recreation with zero behavior change.

**Issue: sponsors on Application fetches data for ALL app registrations**

GitHub Copilot: The root cause is that `RelatedPropertyNames` generates a **per-entity sub-API call** — one `GET /applications/{id}/sponsors` for every app registration in the tenant. The fix is to move `sponsors` into an inline `$expand` on the `QueryString` instead, collapsing N+1 calls into 1.

**In export-tenant.config.psd1, for the `Application` entry:**

```powershell
# Before
QueryString = '$top=999'
RelatedPropertyNames = @('sponsors')

# After
QueryString = '$top=999&$expand=sponsors($select=id)'
RelatedPropertyNames = @()
```

This is exactly the same pattern `ServicePrincipal` already uses for `appRoleAssignments` — `$expand=appRoleAssignments` is embedded in its `QueryString` rather than being a `RelatedPropertyName`. The sponsors data comes back inline with each app in the single listing call instead of requiring a fan-out.

For non-blueprint apps, `sponsors` will simply return an empty array, which has no cost. The spec's own query (`GET /applications/microsoft.graph.agentIdentityBlueprint?$expand=sponsors($select=id)`) is exactly this shape but scoped to the cast collection — the inline `$expand` in the generic `/applications` call produces the same column structure in the DB table, so test queries don't need to change.

alexandair (approved):
LGTM

PR #1249, opened 2026-05-28
Fix Sankey crash when device compliance counts are null

### LINKED ISSUE #1235: ZTA 2.3.0 completes successfully with clean export/database/test logs, but generated HTML report crashes when renderering
ZeroTrustAssessment 2.3.0 completes successfully with clean export, database, and test progress logs, but the generated HTML report crashes when rendering in the browser.

The assessment appears to complete successfully. The export phase completes with no failed jobs, the database phase completes successfully, and the test phase shows all tests completing with `Pass` results. However, when opening `ZeroTrustAssessmentReport.html`, the report fails to render and displays an unexpected application error.

Image

An older Zero Trust Assessment report generated a few months ago still opens successfully in the same browser on the same machine.

dgoldman-msft:
Fixes #1235 - HTML report crashes on tenants where certain device compliance combinations have no data. Measure-Object -Sum on an empty pipeline returns null, which was serialized as 'value: null' in Sankey link JSON, causing D3's Sankey renderer to crash.

Added ?? 0 null coalescing to all Windows and mobile compliance/ non-compliance Measure-Object -Sum calls in Get-DesktopDevicesSummary and Get-MobileSummary.

### DIFF
--- a/src/powershell/private/tenantinfo/devices/Add-ZtDeviceOverview.ps1
+++ b/src/powershell/private/tenantinfo/devices/Add-ZtDeviceOverview.ps1
@@ -79,14 +79,14 @@ order by operatingSystem, trustType, isCompliant
         $windowsEntraRegistered = ($windowsResults | Where-Object { $_.trustType -eq 'Workplace' } | Measure-Object -Property count -Sum).Sum ?? 0
 
         # Windows compliance by join type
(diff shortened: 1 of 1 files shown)


astaykov (approved):
Looks good. Thanks for this fix.

PR #1265, opened 2026-06-02
AI - 61011 - Require users to use Microsoft Entra ID auth to interact with agents

Manoj-Kesana:


src/powershell/tests/Test-Assessment.61011.ps1
```diff
@@ -46,130 +46,95 @@ function Test-Assessment-61011 {
...
-        }
+    try {
+        $agentBlueprints = @(Invoke-DatabaseQuery -Database $Database -Sql $sqlQ2)
     }
-    if (-not $agentBlueprints) {
-        $agentBlueprints = Invoke-ZtGraphRequest `
-            -RelativeUri 'applications/microsoft.graph.agentIdentityBlueprint' `
-            -ApiVersion beta `
-            -Select @('id', 'appId', 'displayName', 'signInAudience')
+    catch {
```
alexandair, line 86: **Q2 failure silently flips all agents to Warning (false Fail)**

```powershell
catch {
    Write-PSFMessage "Failed to query agent identity blueprints: $_" -Tag Test -Level Warning
    $agentBlueprints = @()   # ← silent fallback to empty
}
```

When Q2 fails, `$blueprintByAppId` is empty. Every agent's `$blueprint` resolves to `$null`. The `if ($blueprint)` block is skipped, both signals remain `$false`, and every agent is added to `$warningAgents`. The test returns Fail with a table full of agents all showing empty blueprint names — with no operator-visible indication that the blueprint data couldn't be fetched. This is a silent false-positive failure.

src/powershell/tests/Test-Assessment.61011.ps1
```diff
@@ -46,130 +46,95 @@ function Test-Assessment-61011 {
...
     try {
-        $interactiveSignIns = Invoke-ZtGraphRequest `
+        $interactiveSignIns = @(Invoke-ZtGraphRequest `
             -RelativeUri 'auditLogs/signIns' `
             -ApiVersion beta `
             -Filter "createdDateTime ge $lookbackDate and signInEventTypes/any(t:t eq 'interactiveUser')" `
             -Select @('id', 'createdDateTime', 'userPrincipalName', 'appId', 'resourceId', 'resourceDisplayName', 'signInEventTypes') `
-            -ErrorAction Stop
+            -ErrorAction Stop)
     }
```
alexandair, line 116: **Q3/Q4 failures silently produce false Fail results**

Same problem for the sign-in queries. Both catch blocks swallow the error into `$q3QueryError`/`$q4QueryError` variables that are **never read again** after assignment:

```powershell
$q3QueryError = $null
...
catch {
    $q3QueryError = $_      # stored here...
    Write-PSFMessage ...
}
# $q3QueryError is never checked anywhere below this point
```

When either sign-in query fails, the corresponding lookup map stays empty, signals are never set to `$true`, and agents are classified as Warning. The operator sees a Fail result with all agents apparently lacking Entra authentication — when really the sign-in data just couldn't be retrieved.

**Fix**: Check `$q3QueryError` and `$q4QueryError` before the assessment loop; if both queries failed, skip the test or surface an explicit data-unavailable message.

src/powershell/tests/Test-Assessment.61011.ps1
```diff
@@ -46,130 +46,95 @@ function Test-Assessment-61011 {
...
-    $agentIdentities = $null
-    if ($Database) {
-        $sqlQ1 = @"
-SELECT
-    id,
-    appId,
-    displayName,
-    agentIdentityBlueprintId,
-    accountEnabled
+    $sqlQ1 = @"
```
alexandair, line 54: **`accountEnabled` selected but never used**

Q1 selects `accountEnabled` but the assessment logic never consults it:

```powershell
$sqlQ1 = @"
SELECT id, appId, displayName, agentIdentityBlueprintId, accountEnabled
...
"@
```

A disabled agent identity has zero sign-in activity by design. It will always have Signal 1 = false and Signal 2 = false, landing in `$warningAgents` and causing a tenant-level Fail. The spec doesn't explicitly say to exclude disabled agents, but reporting a disabled agent as "no Entra auth evidence" is misleading. Either filter disabled agents out of Q1, or add a note in the output table row distinguishing them.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.61011.md
```diff
@@ -0,0 +1,12 @@
...
+
+**Remediation action**
+
+- [Authenticate users in interactive agents](https://learn.microsoft.com/en-us/entra/agent-id/identity-platform/interactive-agent-authenticate-user)
+- [Request delegated user authorization for interactive agents](https://learn.microsoft.com/en-us/entra/agent-id/identity-platform/interactive-agent-request-user-authorization)
+- [Agent users in Microsoft Entra Agent ID](https://learn.microsoft.com/en-us/entra/agent-id/identity-platform/agent-users)
+- [Microsoft Entra Agent ID overview](https://learn.microsoft.com/en-us/entra/agent-id/overview)
```
alexandair: Remediation link 4 in the `.md` points to `https://learn.microsoft.com/en-us/entra/agent-id/overview`, but the spec's remediation resource 4 lists `https://learn.microsoft.com/en-us/entra/agent-id/what-is-microsoft-entra-agent-id`. These should match.

src/powershell/tests/Test-Assessment.61011.ps1
```diff
@@ -0,0 +1,268 @@
...
+
+    if (-not $agentIdentities -or $agentIdentities.Count -eq 0) {
+        Add-ZtTestResultDetail -SkippedBecause NotApplicable
+        return
+    }
+
+    # Q2: Enumerate all agent identity blueprints from the exported database
+    Write-ZtProgress -Activity $activity -Status 'Getting agent identity blueprints (Q2)'
+    $sqlQ2 = @"
+SELECT id, appId, displayName, signInAudience
```
alexandair: **`signInAudience` selected in Q2 but never read**

```sql
SELECT id, appId, displayName, signInAudience
FROM main.Application
WHERE "@odata.type" = '#microsoft.graph.agentIdentityBlueprint'
```

`signInAudience` is fetched but nothing in the code ever accesses `$blueprint.signInAudience` or `$agent.signInAudience`. Drop it from the SELECT.

src/powershell/tests/Test-Assessment.61011.ps1
```diff
@@ -0,0 +1,268 @@
...
+- Total agent identities evaluated: {2}
+- Agent identities with warning: {3}
+
+'@
+
+        $mdInfo = $formatTemplate -f $agentPortalUrl, $tableRows, $totalAgentCount, $warningCount
+        }
+    } # end if testResultMarkdown not already set
+
+    $testResultMarkdown = $testResultMarkdown -replace '%TestResult%', $mdInfo
```
alexandair, line 259: **`$mdInfo` undefined in error path**

When the both-fail branch is taken, `$testResultMarkdown` is already set and the outer `if ([string]::IsNullOrEmpty($testResultMarkdown))` block is skipped — meaning `$mdInfo` is never defined. The final line:

```powershell
$testResultMarkdown = $testResultMarkdown -replace '%TestResult%', $mdInfo
```

...uses an undefined `$mdInfo` (PowerShell coerces it silently to empty string). This works today but will break under `Set-StrictMode -Version Latest` and is fragile. Add `$mdInfo = ''` just before the `if ([string]::IsNullOrEmpty($testResultMarkdown))` block.

src/powershell/tests/Test-Assessment.61011.ps1
```diff
@@ -0,0 +1,268 @@
...
+        }
+    }
+    #endregion Data Collection
+
+    #region Assessment Logic
+    $passed = $false
+    $testResultMarkdown = ''
+    $warningAgents = [System.Collections.Generic.List[PSCustomObject]]::new()
+
+    if ($q3QueryError -and $q4QueryError) {
```
alexandair: **Partial sign-in query failure still produces silent false Fail**

The guard is:
```powershell
if ($q3QueryError -and $q4QueryError) {
    $testResultMarkdown = "❌ Unable to evaluate..."
}
else {
    foreach ($agentIdentity in $agentIdentities) { ... }
    $passed = $warningAgents.Count -eq 0
}
```

This only handles the case where **both** queries fail. If only Q3 fails, `$interactiveSignInsByBlueprintObjectId` is empty — Signal 2 can never be `$true` for any agent. Agents with interactive sign-in evidence only (Signal 2 agents) are incorrectly classified as Warning and show up in the failure table with no indication the data was unavailable. Same issue if only Q4 fails (Signal 1 permanently false).

Fix: test each error independently. When Q3 fails but Q4 succeeded, note in the output that Signal 2 data is unavailable; similarly for Q4. Alternatively, treat either query failure as a reason to skip or add a visible caveat row to the table.

alexandair (changes requested):

alexandair (approved):
LGTM

PR #1276, opened 2026-06-04
Updated Error logging on sentinal workspace

Manoj-Kesana:


src/powershell/private/tests-shared/Get-SentinelWorkspaceData.ps1
alexandair, line 57: **Q2 is not scoped to Q1's enabled subscriptions**

Q1 explicitly filters for `state =~ 'Enabled'`, but Q2:
```kql
resources
| where type =~ 'microsoft.operationalinsights/workspaces'
| project workspaceName=name, workspaceId=id, resourceGroup, subscriptionId
```
returns workspaces from **all accessible subscriptions**, including disabled ones. When a workspace's `subscriptionId` is absent from `$subscriptionNameById`, the hashtable lookup silently returns `$null`:

```powershell
SubscriptionName = $subscriptionNameById[$workspace.subscriptionId]  # $null for disabled-sub workspaces
```

Downstream, `Get-SafeMarkdown -Text $null` in the caller will produce an empty column in the report. More importantly, the Sentinel Q3 check will also run against those workspaces, which is likely unintended.

The old single-query approach had the same gap (`leftouter` join with null `subscriptionName`), so this isn't a regression — but the split-query refactor was a good opportunity to close it. Suggested fix: after Q2, filter to only workspaces whose subscriptionId is in `$subscriptionNameById`:

```powershell
$allWorkspaces = @($allWorkspaces | Where-Object { $subscriptionNameById.ContainsKey($_.subscriptionId) })
```

When you fix the above issue, you will fix this one, too:

**`'NoWorkspaces'` can be bypassed by disabled-subscription workspaces (consequence of Issue 1)**

Because Q2 includes workspaces from disabled subscriptions, if Q1 returns enabled subscriptions with zero workspaces but a disabled subscription happens to have workspaces, Q2 returns a non-empty result — the `'NoWorkspaces'` guard is skipped and the caller runs the full assessment on workspaces that have `SubscriptionName = $null`. Resolving Issue 1 fixes this automatically.

alexandair (changes requested):

src/powershell/private/tests-shared/Get-SentinelWorkspaceData.ps1
Manoj-Kesana: Resolved

alexandair (approved):
LGTM

PR #1284, opened 2026-06-04
fix: prevent Sankey crash on empty device categories

astaykov:
First Sankey split PR: backend null-coalescing and defensive link filtering to prevent Sankey crashes.


astaykov:
Closing per request; keeping branch split only.

PR #1285, opened 2026-06-04
refactor: use generic empty state message in Sankey component

astaykov:
Second Sankey split PR: isolates the empty-state message refinement.


astaykov:
Closing per request; keeping branch split only.

PR #1299, opened 2026-06-10
fix: added skip scenario to Test 25392 when there are no private access connectors installed

### LINKED ISSUE #1264: Test-25392 shows passed with no private access connectors installed. It should be "skipped"
The test 25392 - `Private Access Connectors are running the latest version` shows a `Passed` status when no private access connectors are detected.
When there are no private access connectors detected, the test should have `Skipped` status with user facing message: `No Private Access connectors were detected.`

Image

praneeth-0000:
closes #1264 

Made changes to test 25392 to show skip scenario when there are no private access connectors installed with user facing message: `No Private Access connectors were detected.`

image


alexandair (approved):
LGTM

PR #1300, opened 2026-06-10
Device - 51001 - Windows Endpoint Privilege Management is configured and assigned

Manoj-Kesana:


alexandair (changes requested):
@Manoj-Kesana Add a per-policy deep link on the policy name column.

alexandair (approved):
LGTM

PR #1309, opened 2026-06-15
Device 51015 - Multi Admin Approval is enabled in Intune to require a second admin to approve sensitive tenant changes

ashwinikarke:


src/powershell/tests/Test-Assessment.51015.ps1
```diff
@@ -0,0 +1,159 @@
...
+            }
+        }
+    }
+    #endregion Data Collection
+
+    #region Assessment Logic
+    # Pass: at least one Q2 policy detail has a non-empty approverGroupIds collection.
+    # Fail: zero policies exist, OR every Q2 detail has an empty approverGroupIds.
+    $passed = @($policyDetails.Values | Where-Object { $_ -and $_.approverGroupIds -and $_.approverGroupIds.Count -gt 0 }).Count -gt 0
+
```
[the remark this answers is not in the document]
ashwinikarke: fixed

src/powershell/tests/Test-Assessment.51015.ps1
```diff
@@ -0,0 +1,175 @@
...
+        else {
+            $policies
+        }
+
+        foreach ($policy in $displayedPolicies) {
+            $policyName = Get-SafeMarkdown $policy.displayName
+            $policyType = $policy.policyType
+            $platform = $policy.policyPlatform
+            $lastModified = Get-FormattedDate $policy.lastModifiedDateTime
+            # Use Q2 detail for approverGroupIds; fall back to 0 if detail fetch failed.
```
alexandair: **Q2 partial failures produce misleading table rows**

When a Q2 call fails for policy `X`, `$policyDetails[$policy.id]` is `$null`, so `$approverGroupCount = 0` and `$rowStatus = '❌ Fail'` for that row. An assessor viewing the table cannot distinguish between "genuinely no approver group" and "API call failed." Only a full Q2 blackout is caught by the guard:

```powershell
# Only guards against ALL Q2 calls failing — partial failures are silent
if ($policies.Count -gt 0 -and $policyDetails.Count -eq 0) { ... }
```

This manifests at [Test-Assessment.51015.ps1](c:\gh\zerotrustassessment\src\powershell\tests\Test-Assessment.51015.ps1#L134-L146): the `else { 0 }` fallback on line 138 collapses a failed Q2 fetch into `$approverGroupCount = 0`, and the `$rowStatus` assignment on line 140 then renders it as `❌ Fail` — identical to a genuine policy with no approver group.

Suggestion: mark rows where `$policyDetails[$policy.id]` is `$null` with a distinct status like `⚠️ Unknown` instead of `❌ Fail`.

src/powershell/tests/Test-Assessment.51015.ps1
```diff
@@ -0,0 +1,175 @@
...
+
+        if ($policies.Count -gt 10) {
+            $tableRows += "| ... ($($policies.Count) total) | | | | | |`n"
+        }
+
+        $formatTemplate = @'
+
+## [Intune Multi Admin Approval access policies]({1})
+
+| Policy name | Policy type | Platform | Approver group count | Last modified | Status |
```
alexandair, line 159: ** Deep links on policy display names**

We prefer deep-linking display names to admin portal pages rather than linking only the table header. If Intune exposes a per-policy URL (e.g., via the policy `id`), the `$policyName` cell could link directly to that policy page. If no reliable per-policy deep link exists, the current header link is acceptable.

alexandair (changes requested):

alexandair (approved):
LGTM

Issue #1230, opened 2026-05-20
Running in DoD Cloud and Signins Fail to Security and Compliance as well as AIP


skirkpatrickMSFT:
<img width="926" height="373" alt="Image" src="https://github.com/user-attachments/assets/c30b5802-e032-491a-96d9-e0490c3d15e1" />

praneeth-0000:
Hi @skirkpatrickMSFT , Thanks for raising this issue. A fix for this issue has been merged in pre-release version. Could you please install the pre-release version and try again?

To test on the pre-release version, follow these steps and then run the assessment:

If you’ve previously run Zero Trust Assessment, uninstall all existing versions:
```
Uninstall-Module ZeroTrustAssessment -Force -AllVersions
```

Install the latest pre-release module:
```
Install-Module ZeroTrustAssessment -AllowPrerelease
```

astaykov:
@skirkpatrickMSFT ,
can you validate with the latest version as instructed.

wshep007:
The preview version did not fix the issue for me.  However, I did find that if I imported the AIP and SPO modules before running the connect-ztassessment it connected me to all the modules.

import-module Microsoft.Online.SharePoint.PowerShell -UseWindowsPowerShell
import-module AIPService -UseWindowsPowerShell
Connect-ztAssessment

astaykov:
Thank you. We are glad that you found a resolution.
We will close this issue for the time being.

PR #1312, opened 2026-06-15
fix: issue-1230 - updated AIP service env for USGovDOD & SecurityCompliance AADAuthEndpointUri

praneeth-0000:
This PR addresses #1230 

**Issue**

Two bugs were found in `Connect-ZtAssessment` affecting USGovDoD connections:

1. **AIP** : The `USGovDoD` environment path passed `AzureUSGovernment2` to `Connect-AipService`, which is not a valid value. Per the [official docs](https://learn.microsoft.com/en-us/powershell/module/aipservice/connect-aipservice?view=azureipps#-environmentname), the only accepted values are `AzureCloud`, `AzureChinaCloud`, and `AzureUSGovernment`, causing the connection to fail.

2. **Security & Compliance** : The `AzureADAuthorizationEndpointUri` values used `/common` for all environments. The [Connect-IPPSSession docs](https://learn.microsoft.com/en-us/powershell/module/exchangepowershell/connect-ippssession?view=exchange-ps#-azureadauthorizationendpointuri) explicitly recommend `/organizations` for enterprise and government tenants.

**Fix**

1. `USGovDoD` now maps to `AzureUSGovernment` (matching `USGov`) when connecting to AIP via `Connect-AipService`.
2. All `AzureADAuthorizationEndpointUri` values in the Security & Compliance environment map updated from `/common` to `/organizations`.

### DIFF
--- a/src/powershell/public/Connect-ZtAssessment.ps1
+++ b/src/powershell/public/Connect-ZtAssessment.ps1
@@ -481,7 +481,7 @@ function Connect-ZtAssessment {
 				$aipEnvironmentName = switch ($Environment) {
 					'China'    { 'AzureChinaCloud' }
 					'USGov'    { 'AzureUSGovernment' }
-					'USGovDoD' { 'AzureUSGovernment2' }
+					'USGovDoD' { 'AzureUSGovernment' }
 					default    { 'AzureCloud' }
 				}
 				if ($ClientId -and $Certificate) {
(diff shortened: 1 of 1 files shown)


astaykov (approved):
Looks good.

alexandair (approved):
@praneeth-0000 
If we don't plan to support MSA accounts even for 'O365Default' environment, the change from `common` to `organizations` is fine.

PR #1314, opened 2026-06-15
Fixed the remaining concrete Sankey producer bug and added one narrow renderer guard.

### LINKED ISSUE #1310: "Unexpected Application Error! Invalid array length" in HTML report
Hello,
i have used several times this tool without particular problems, but with the latest versions i am getting the attached error.
With v 2.3.0 and 2.4.0 i can have a readable report only if select 1 pillar (identity in my cases), otherwise i get the error.

Before i used v 2.1.0 and 2.2.0 and i had the 2 version installed at the same time, now i removed every version, the folder was empty and i only have the 2.4.0 and PowerShell 7.6.2.

During the execution i only had this warning:
WARNING: [09:01:36][Wait-ZtTest] Error processing Test 21814. | Response status code does not indicate success: NotFound (Not Found).
"Privileged accounts are cloud native identities"

Image

astaykov:
This fix should close #1310

`src\powershell\private\tenantinfo\Add-ZtOverviewCaDevicesAllUsers.ps1` now sums cnt values before casting. The old code did `($Results | Where-Object isManaged).cnt -as [int]`, which turns the two managed rows into $null instead of the total. That was the likeliest remaining way to emit a broken homepage Sankey graph.

`src\report\src\components\nivo\sankey.tsx` now also drops links whose source or target is not in the declared node list, on top of the existing null/non-positive filtering. That keeps malformed graph topology away from d3-sankey, which is where the Invalid array length error originates.

`code-tests\commands\Add-ZtOverviewCaDevicesAllUsers.Tests.ps1` covers the exact issue shape from #1235/#1272: two managed buckets plus one unmanaged bucket must produce a valid User sign in -> Managed total.

### DIFF
--- a/code-tests/commands/Add-ZtOverviewCaDevicesAllUsers.Tests.ps1
+++ b/code-tests/commands/Add-ZtOverviewCaDevicesAllUsers.Tests.ps1
@@ -1,0 +1,57 @@
+Describe "Add-ZtOverviewCaDevicesAllUsers" {
+	BeforeAll {
+		$here = $PSScriptRoot
+		$srcRoot = Join-Path $here "../../src/powershell"
+
+		if (-not (Get-Command Write-PSFMessage -ErrorAction SilentlyContinue)) {
+			function global:Write-PSFMessage {
+				param($Level, $Message, $Tag)
+			}
+		}
+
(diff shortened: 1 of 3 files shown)


alexandair (approved):
LGTM

Issue #586, opened 2025-11-19
Customize Assessment Test Results


FriedrichWeinmann:
It would be great if we could provide a configuration file with customizations to the test results, that show up in the final dashboard.

For Example:

+ Comments. A simple comment that shows up in the finding with no further change.
+ Result override: Override the result ...
  + Specify a new result
  + Provide a Justification text, _why_ the result was overridden
  + Specify an override condition, so that a change in findings may retrigger a fail

This might also be extended to provide a default system to provide Test-specific settings. The settings would then be something Tests _could_ implement (the settings will not matter for tests, that do not implement any).

> Example Configuration file

Here a quick example on how such a configuration file might look:

```powershell
@{
	DisabledTests = @(
		21770
		21771
		21772
		21773
		21774
	)

	Config = @{
		'21780' = @{
			Comment = 'Example Comment'
		}
		'21781' = @{
			Override = $true
			NewStatus = 'Compensated'
			Justification = 'The requirements this Test verifies are met with third party tool XYZ, which the assessment cannot detect.'
		}
		'21849' = @{
			Config = @{
				AppsToIgnore = @(
					'5c963c5b-b3f4-4eac-8a95-65b1abe6f20c' # Break Glass Tenant Restore Application
				)
			}
		}
	}
}
```

Any thoughts on the matter?
I wouldn't mind implementing the overall framework to make this happen, but do we want to support this?

PR #1317, opened 2026-06-16
Caching Rework, Timeouts & Memory Improvements

FriedrichWeinmann:
Alright, this took a bit longer than expected, but it's finally there:
A major update to how data is persisted within memory to help with large tenant setups and require less configuration. Should work better out-of-the-box.

## Changes

+ Upd: Raised minimum PSFramework version, due to feature usage
+ Upd: Migrated test timeouts for parallelized tests to Runspace Workflow timeouts, enabling activity-based timeouts
+ Upd: Added support to retry failed tests
+ Upd: Migrated Caches to PSFCache - limiting data retention by count and time
+ Upd: Added soft-limit for Test Statistics messages
+ More minor tweaks

### Test Timeout update

This update only applies to the tests executed in the background.
It uses the latest PSFramework update to the Runspace Workflows to support timeouts. The key difference to the previous implementation is that these timeouts can be based on _activity_, rather than "since the start". Activity is measured whenever you use `Write-PSFMessage`.

As the tests running in the main runspace cannot use this feature, the test timeout is kept at the previous one hour by default.
A future update will address that if possible.

### Retry Failed Tests

This update only applies to the tests executed in the background.
It uses the latest PSFramework update to the Runspace Workflows to support retries on test failure. The number of retries can be defined via configuration and defaults to 0 extra attempts.
A future update will support customizing this count on a per-test basis.

> Note: In the implementation I retry once more than specified. This is so a last round of the retry validation can process the "test failed" handling.

### Migrated Caches to PSFCache

We cache all requests to Graph or Azure unless configured otherwise.
This can add a huge load in large tenants.

The new system will automatically expire values too old or when too many values are cached.
Both Azure and Graph request caches default to 30 minutes of age or 50,000 entries, whichever happens first. All limits can be customized via configuration.

### Added soft-limit for Test Statistics messages

When you run `Get-ZtTestStatistics`, you can collect information about the last test execution.
This includes all log messages generated.

Since some of our tests have scaling message output based on tenant size, this could lead to lots of memory being used by these messages.
With this new configuration setting, once too many messages are logged for a test - 1024 by default - the data retained is implified significantly to reduce memory overhead, at the loss of some data.

## Warning: Untested

Due to technical limitations I am currently unable to test this code on my own. Given the pressing nature of the large tenant performance, I chose to push this now, rather than wait until I can make the time to restore my own testing capability.

**Do not merge right away without testing it first!!!**

## Thoughts on Cache Changes & Paged Requests

As part of the design change, I considered whether this change will lead to issues with our habit of caching paged requests separately, one cache entry per page.
This might theoretically lead to issues with partial cache invalidation.

In case you want to consider the same issues, here's the potential issues my mind went to and my conclusions on them:

> When only part of a paged request is invalidated from cache, will results become inconsistent?

No.
Whether by max items or by lifetime, cache content is invalidated from the oldest entries.
In this situation, the first request is never going to be from cache, its "next page" link will have an entirely different continuation token and thus not read from cache at all.

The only situation where an issue might occur is when reconfiguring cache limits while it is reading from cache.
A situation that is purely theoretical, since the cache is only used during the full processing where the console is locked.

> What about concurrency messing with paged cache entries?

Theoretically, other runspaces could add new data too fast and - while one runspace is reading cached pages, others could add new entries and expire the next page from the cache.
The continuation token is unlikely to still be valid and the read fails.

Theoretically an issue, in practice exceedingly unlikely.
Reads from cache are _far_ faster than rest requests, which means enough other runspaces would need to add new cache entries in parallel to outpace that.
This would not be an issue in almost any conceivable scale configuration.


astaykov (changes requested):
This is not a stable PR. 
Neither Import-Module, nor Connect-ZtAssessment, nor Invoke-ZtAssessment complete successfully. The main issue is a warning:

```powershell
PS C:\tdb\zerotrustassessment> Import-Module .\src\powershell\ZeroTrustAssessment.psd1 -Force

Resolving 7 dependencies...
    ✅ Module PSFramework found (v1.14.449).
# trimmed for clarity

Asserting MSAL loading order for dependencies...
    ✅ Loading Microsoft.IdentityModel.Abstractions.dll
# trimmed for clarity
WARNING: [12:38:43][ZeroTrustAssessment.psm1] Failed to import file C:\tdb\zerotrustassessment\src\powershell\scripts\variables.ps1 | Cannot convert argument "Lifetime", with value: "PSFramework.Configuration.Config", for "SetLifetime" to type "PSFramework.Parameter.TimeSpanParameter": "Cannot convert value "PSFramework.Configuration.Config" to type "PSFramework.Parameter.TimeSpanParameter". Error: "Could not interpret PSFramework.Configuration.Config""

PS C:\tdb\zerotrustassessment> Connect-ZtAssessment -TenantId abc.com
🔑 Authentication to Graph, Azure, AipService, ExchangeOnline, SecurityCompliance, SharePointOnline.
During the next steps, you may be prompted to authenticate separately for several services.

Connecting to Microsoft Graph
   ✅ Connected
WARNING: [12:41:38][Connect-ZtAssessment] Could not resolve the initial tenant domain from Graph. EXO, S&C, and SPO CBA connections may fail.
WARNING: [12:41:38][Test-ZtContext] Unable to verify user roles | Cannot index into a null array.

```

Ref. automated validation run (but I also tested this on a clean VM): https://github.com/microsoft/zerotrustassessment/actions/runs/27681160160/job/81868692154

FriedrichWeinmann:
Thanks for catching that! Fixed.

src/powershell/private/tests/Start-ZtTestExecution.ps1
```diff
@@ -86,27 +85,74 @@
...
+			$result.TimedOut = $_ -match 'Workitem timed out'
+
+			# Let's fail this
+			$result.End = Get-Date
+			$result.Duration = $result.End - $result.Start
+
+			Write-ZtTestError -Test $this -Result $result -ErrorRecord $_
+			Write-PSFMessage -Message "Processing test '{0}' - Concluded" -StringValues $Test.TestID -Target $Test -Tag end
+			Write-ZtTestFinish -Result $result -PreviousMessages $global:msgSoFar[$_.TestID] -Test $this -LogsPath $logsPath
+		}
```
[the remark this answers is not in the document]
FriedrichWeinmann: The finding on the `Get-ZtTestResult` parameter name is correct and fixed for the next commit. As is the missing `$Test` variable. It is wrong about `$this` not existing, which is something PSFramework adds to the retry condition block and contains the Test configuration object.

src/powershell/private/tests/Invoke-ZtTests.ps1
```diff
@@ -71,7 +71,7 @@
 		[TimeSpan]
 		$Timeout = '1.00:00:00',
 
-		[int]
+		[PSFTimeSpan]
 		$TestTimeout = 60
 	)
```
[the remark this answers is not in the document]
FriedrichWeinmann: The finding is correct if inconsequential, as the default value should never be used, as we always pass through the value from Invoke-ZtAssessment. Still, it is unclean and updated for the upcoming Commit. Also fixed the comment to reflect the actual behavior.
Note Copilot: Your suggested update would have removed a syntax-critical parenthesis.

src/powershell/public/Invoke-ZtAssessment.ps1
```diff
@@ -178,14 +174,14 @@ function Invoke-ZtAssessment {
...
+		[PSFTimeSpan]
 		$Timeout = '1.00:00:00',
 
 		# Maximum time in minutes a single test is allowed to run. Defaults to 60 minutes. Set to 0 to disable.
 		# For Data pillar tests, timeout is best-effort because some external modules/remoting
 		# operations cannot be deterministically hard-stopped from within the current process.
-		[int]
-		$TestTimeout = [math]::Floor((Get-PSFConfigValue -FullName 'ZeroTrustAssessment.Tests.Timeout' -Fallback ([timespan]::FromMinutes(60))).TotalMinutes),
+		[PSFTimeSpan]
+		$TestTimeout = (Get-PSFConfigValue -FullName 'ZeroTrustAssessment.Tests.Timeout' -Fallback ([timespan]::FromMinutes(60))),
```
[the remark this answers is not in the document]
FriedrichWeinmann: Correct finding, updated the help, both directly above and in the CBH at the top of the file for the upcoming commit.

src/powershell/private/tests/Get-ZtTestResult.ps1
```diff
@@ -0,0 +1,56 @@
...
+		Retrieve the test result object for the specified test.
+		This will - on first call - generate a new, mostly empty object.
+		On subsequent calls - during retries or when considering on whether to retry at all - it will return the same object, which presumably has been updated since.
+
+	.PARAMETER Test
+		The test object for which we need a result dataset
+
+	.EXAMPLE
+		PS C:\> Get-ZtTestResult -TestID $Test
+
```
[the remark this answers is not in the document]
FriedrichWeinmann: Correct and fixed for the upcoming commit

src/powershell/private/tests/Invoke-ZtTestParallel.ps1
```diff
@@ -0,0 +1,118 @@
...
+
+	.EXAMPLE
+		PS C:\> Invoke-ZtTestParallel -Test $_ -Database $global:database -LogsPath $logsPath
+
+		Executes the current test with the globally cached database connection and writes a log file.
+
+	.EXAMPLE
+		PS C:\> Invoke-ZtTestParallel -Test $_ -Database $global:database -LogsPath $logsPath -TestTimeout ([timespan]::FromMinutes(30))
+
+		Executes the test with a 30-minute timeout.
```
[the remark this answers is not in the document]
FriedrichWeinmann: Correct, was a copy&paste leftover when cloning it from Invoke-ZtTest. Dropped the second example as no longer relevant.

src/powershell/scripts/variables.ps1
```diff
@@ -9,8 +9,8 @@ if ($prerelease) { $moduleVersion = "$moduleVersion-$prerelease" }
 
 $script:__ZtSession = @{
 	# A DCO dictionary is the same threadsafe dictionary across all runspaces, allowing parallelized checks to write results to the same store safely
-	GraphCache   = Set-PSFDynamicContentObject -Name "ZtAssessment.GraphCache" -Dictionary -PassThru
-	AzureCache   = Set-PSFDynamicContentObject -Name "ZtAssessment.AzureCache" -Dictionary -PassThru
+	GraphCache   = Set-PSFDynamicContentObject -Name "ZtAssessment.GraphCache" -Cache -PassThru
+	AzureCache   = Set-PSFDynamicContentObject -Name "ZtAssessment.AzureCache" -Cache -PassThru
```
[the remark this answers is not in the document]
FriedrichWeinmann: Good catch, updated it for the upcoming commit

alexandair:
@FriedrichWeinmann 

`variables.ps1` switches `GraphCache`/`AzureCache` from `-Dictionary` to `-Cache` (`PSFramework.Caching.CacheMemoryConcurrent`). But the cache‑hit checks in Invoke-ZtGraphRequestCache.ps1 and Invoke-ZtAzureRequestCache.ps1 gate reads on `.ContainsKey()`:

```powershell
$isInCache = $script:__ZtSession.GraphCache.Value.ContainsKey($cacheKey)
if (... -and $isInCache -and $isMethodGet) { $results = ...Value[$cacheKey]; if ($results) { return $results } }
```

PSFramework 1.14.450's `CacheBase.ContainsKey` has **inverted logic** (verified in source and empirically on the installed module):

```csharp
public override bool ContainsKey(object key) {
    if (!base.ContainsKey(key)) return false;
    return ((CachedData)base[key]).IsExpired;   // BUG: should be !IsExpired
}
```

So for a *fresh* entry, `ContainsKey` returns **`False`** (and `True` only once expired). I confirmed empirically: after `$c[$key]=value`, `ContainsKey=False` while `Count=1`, `.Keys` contains the key, and the indexer returns the value.

**Impact:** `$isInCache` is always `$false` → the cache-read branch is never entered → **every Graph/Azure GET bypasses the cache and re-hits the API.** Writes still happen, so the cache fills with entries that are never read. This is a regression from the previously-working dictionary cache and directly contradicts the PR's caching/throttling/perf goals.

**Fix options:**
1. Stop relying on `ContainsKey()`; read via the indexer and null-check (the cache doesn't store nulls — `GetCacheNull()=False` — so this is safe and matches the existing `if ($results)` guard):
   ```powershell
   if (-not $cacheBlocked -and -not $DisableCache -and -not $isBatch -and $isMethodGet) {
       $results = $script:__ZtSession.GraphCache.Value[$cacheKey]
       if ($results) { Write-PSFMessage "Using graph cache: $cacheKey" -Level Debug; return $results }
   }
   ```
2. Or fix `CacheBase.ContainsKey` upstream (add the missing `!`) and pin to that build — relevant since the PR author maintains PSFramework.
3. Or revert the caches to `-Dictionary`.

src/powershell/private/tests/Start-ZtTestExecution.ps1
```diff
@@ -86,27 +85,74 @@
...
+
+			$result.TimedOut = $_ -match 'Workitem timed out'
+
+			# Let's fail this
+			$result.End = Get-Date
+			$result.Duration = $result.End - $result.Start
+
+			Write-ZtTestError -Test $this -Result $result -ErrorRecord $_
+			Write-PSFMessage -Message "Processing test '{0}' - Concluded" -StringValues $this.TestID -Target $this -Tag end
+			Write-ZtTestFinish -Result $result -PreviousMessages $global:msgSoFar[$_.TestID] -Test $this -LogsPath $logsPath
```
alexandair: **Wrong variable in retry condition bloats failed-test memory**

In the condition scriptblock, Start-ZtTestExecution.ps1:127:

```powershell
Write-ZtTestFinish -Result $result -PreviousMessages $global:msgSoFar[$_.TestID] -Test $this -LogsPath $logsPath
```

I confirmed from PSFramework's `WorkerRetryCode` that inside the condition **`$_` is the ErrorRecord and `$this` is the test**. So `$_.TestID` is `$null` → `$global:msgSoFar[$null]` → `PreviousMessages = $null`. In `Write-ZtTestFinish`, `Get-PSFMessage | Where { $_ -notin $null }` then keeps **every message accumulated in that worker runspace** (from all prior tests) and attributes them to this one failed/timed-out test's statistics and log. Should be `$this.TestID` (the sibling `process` block correctly uses `$_.TestID` because there `$_` *is* the test).

src/powershell/scripts/configuration.ps1
```diff
@@ -1,9 +1,28 @@
...
+Set-PSFConfig -Module ZeroTrustAssessment -Name 'ThrottleLimit.Export' -Value 5 -Initialize -Validation integerpositive -Description 'Maximum number of data collectors processed in parallel'
+
+# Test Settings
+Set-PSFConfig -Module ZeroTrustAssessment -Name 'ThrottleLimit.Tests' -Value 5 -Initialize -Validation integerpositive -Description 'Maximum number of tests processed in parallel'
 Set-PSFConfig -Module ZeroTrustAssessment -Name 'Tests.Timeout' -Value '1h' -Initialize -Validation timespan -Description 'Maximum time a single test is allowed to run before it is stopped. Defaults to 1 hour. Set to 0 to disable the timeout. Tests that exceed this limit are recorded as timed out and execution continues with the next test.'
+Set-PSFConfig -Module ZeroTrustAssessment -Name 'Tests.TimeoutType' -Value 'Idle' -Initialize -Description 'How the timeout for test execution is measured: Idle = "Since last activity", Start = "Since the Test was started"'
```
alexandair: **Medium — Parallel timeout semantics silently changed (Idle vs wall-clock)**

Parallel tests now use the V2 worker timeout with `Tests.TimeoutType = 'Idle'` (measured from last activity / last `Write-PSFMessage`), while the sync/Data path still uses the wall-clock `TimeoutHelper`. A long-running test that logs periodically can keep resetting the idle window and **never hit the configured cap**. The old parallel path enforced a from-start timeout. This is a real behavior change worth calling out in the PR/docs; use `Start` if a hard per-test cap is intended.

alexandair:
**Observations**

- **Fragile retry contract:** PSFramework's `ShouldRetry` decides retry from `IsTrue(result[0])` — the *first object the condition emits*. It works today only because every helper on the no-retry path (`Write-ZtTestError`, `Write-ZtTestFinish`, `Write-ZtTestLog`, `Write-ZtTestProgress`) carefully suppresses output (I checked each). Add an explicit `return $false` at the end of the condition to harden against any future helper that emits an object (which would cause spurious retries).
- **No test coverage** for the reworked path (`Get-ZtTestResult`, `Invoke-ZtTestParallel`, the retry condition, `Write-ZtTestFinish`); repo convention asks for tests on behavior changes.
- **`$script:_testCache` (Get-ZtTestResult) is never cleared** — bounded in practice (fresh runspaces per run; same objects also held by `TestStatistics`), so minor, but worth a comment/clear hook.
- **`Tests.TimeoutType` config has no `-Validation`** — mitigated by the runtime enum check + `Idle` fallback in `Start-ZtTestExecution`.
- **Doc staleness:** `Start-ZtTestExecution`'s `.PARAMETER TestTimeout` still says "Passed through to Invoke-ZtTest," but it now configures the worker timeout.
- **Timeout detection by string match** (`$_ -match 'Workitem timed out'`) couples to PSFramework's (non-localized) exception text; acceptable, but a comment or `[System.TimeoutException]` guard would be more robust.

alexandair (changes requested):
@FriedrichWeinmann Please, address my feedback.

FriedrichWeinmann:
Hi Alexandair, thank you for the thorough review.
Especially appreciating how you followed up on that upstream bug to where it's happening.

I can't disagree with any of it, will try to fit in remediations / adjustments as soon as I am able to.

FriedrichWeinmann:
Alright, got around to wrap up the findings and address the points:

## Findings / Issues

> Cache check issue with `ContainsKey()`

Fixed the bug in PSFramework (and added extensive tests for it to prevent regression), raised the required version.
But ... the pattern in the Invoke-Zt*RequestCache commands was problematic with the new invalidating cache: The cache might get invalidated between the check and the retrieval. Thus I went with your proposed option 1, rather than keep relying on ContainsKey.

> Fragile retry contract

I somewhat disagree on your arguments here. As you mentioned I _did_ make sure my commands did not return output.
I have however added explicit null-assignments in the condition block in case somebody later sneaks in some accidental returns in the helpers or the commands they call.
I am curious what the proposed explicit `return $false` is supposed to do though, given that a leaked output object would already make the condition evaluate as $true.

> No Test Coverage

Added Tests for the three mentioned commands:

- Get-ZtTestResult
- Invoke-ZtTestParallel
- Write-ZtTestFinish

> $script:_testCache (Get-ZtTestResult) is never cleared

Correct. I've added a cleanup in `Write-ZtTestFinish` and added a note in `Get-ZtTestResult`'s description that mentions it.

> Tests.TimeoutType config has no -Validation

Added a validator in the PSFramework update and updated the setting to now validate.

> Doc staleness: Start-ZtTestExecution

Absolutely correct. Fixed.

> Timeout detection by string match

Correct. There were some historic issues with wrapped error records & exceptions that I resolved before shipping the PSFramework feature. It's reliable now, but still had me lacking confidence in that. Fixed.

> Wrong variable in retry condition bloats failed-test memory

Thanks for catching that, fixed.

> Parallel timeout semantics silently changed (Idle vs wall-clock)

`This is a real behavior change worth calling out in the PR`

Which is why this is literally the first section under the index of changes.

So, here's my thoughts on the overall topic:
The ZeroTrustAssessment tries to give a comprehensive analysis of your tenant's security configuration. This means precision and detail are important, speed is secondary to that (but still highly desirable, of course). But this means, that cancelling a Test before it's done is an issue:

- It makes us lose the information, which is bad from a goals perspective.
- If the Test is actually hanging, we risk perpetually, _never_ reaching a result (which is obviously even worse).

A linear "1 hour" timeout was very generous and a failsafe against the hang. But it has another issue: Large tenants can have individual Tests take excessively longer, simply by having more data to delve through. They might hit the limits while still progressing.
This means those ZTA users will now have to figure out a safe boundary or accept _never_ finishing a specific Test. Which also means it does not reliably work for them "out-of-the-box" - they _have_ to deal with it.
And when they raise the timeout and have a true hang, they wait longer, since the wait is for all Tests.

This is where the Idle Timeout comes in:
It allows us to determine true hangs as opposed to legitimately long-running Tests. Which means healthy but slow tests can complete without requiring interaction with the settings.

## Future musings

Some of the "making this behavior more visible to the end user" is planned for an upcoming PR on my roadmap:
Implementation of Config files for the ZTA to more conveniently customize execution. Some of my plans for that can be read here: https://github.com/microsoft/zerotrustassessment/issues/586
That future PR is also going to include documented example config files and go into more of the performance-related settings.

src/powershell/private/tests/Start-ZtTestExecution.ps1
FriedrichWeinmann: fixed

src/powershell/scripts/configuration.ps1
FriedrichWeinmann: > This is a real behavior change worth calling out in the PR

The first detail entry under the index of changes here in the PR.
The config settings are also documented.
My philosophical position on this point down below in the main thread :)

src/powershell/public/Invoke-ZtAssessment.ps1
```diff
@@ -178,14 +174,14 @@ function Invoke-ZtAssessment {
...
+		[PSFTimeSpan]
 		$Timeout = '1.00:00:00',
 
-		# Maximum time in minutes a single test is allowed to run. Defaults to 60 minutes. Set to 0 to disable.
+		# Maximum time a single test is allowed to run. Defaults to 60 minutes. Set to 0 to disable.
 		# For Data pillar tests, timeout is best-effort because some external modules/remoting
 		# operations cannot be deterministically hard-stopped from within the current process.
-		[int]
-		$TestTimeout = [math]::Floor((Get-PSFConfigValue -FullName 'ZeroTrustAssessment.Tests.Timeout' -Fallback ([timespan]::FromMinutes(60))).TotalMinutes),
+		[PSFTimeSpan]
```
alexandair, line 184: **Numeric `-TestTimeout` values now use seconds, not documented minutes**

Invoke-ZtAssessment.ps1 changed from `[int]` to `PSFTimeSpan`.

A formerly valid call such as `Invoke-ZtAssessment -TestTimeout 60` now becomes **60 seconds**, not 60 minutes. A JSON numeric configuration value such as `"TestTimeout": 60` has the same result; a string `"60"` becomes **60 days**. The value is then passed unchanged through Invoke-ZtTests.ps1:74-75.

This can cause normally running tests to time out after one minute. Preserve minute-based numeric input or update the parameter/docs/config contract with a compatible conversion.

src/powershell/private/tests/Invoke-ZtTestParallel.ps1
```diff
@@ -0,0 +1,113 @@
...
+	}
+	process {
+		Write-PSFMessage -Message "Processing test '{0}'" -StringValues $Test.TestID -Target $Test -Tag start
+
+		# Update progress dashboard: map this runspace to this test and mark as starting
+		$runspaceId = [runspace]::DefaultRunspace.InstanceId.ToString()
+		$script:__ZtSession.ProgressState.Value["rs_$runspaceId"] = $Test.TestID
+		Update-ZtProgressState -WorkerId $Test.TestID -WorkerName $result.DisplayName -WorkerStatus 'Running' -WorkerDetail 'Starting test...'
+
+		# Check if the function exists and what parameters it has
```
alexandair, line 57: `Invoke-ZtTestParallel` can throw while resolving the command at Invoke-ZtTestParallel.ps1:58-61, before it assigns `$result.Start` in Invoke-ZtTestParallel.ps1.

1. At Invoke-ZtTestParallel.ps1, the worker may fail immediately if the test command cannot be found.
2. The result’s `Start` timestamp is only assigned later, inside Invoke-ZtTestParallel.ps1:87-93.

So this failure sequence is possible:

```text
Test command is missing
  → throws before `result.Start = Get-Date`
  → workflow error handler runs
  → attempts `result.End - result.Start`
  → `Start` is null, causing a new error
```

src/powershell/private/tests/Invoke-ZtTest.ps1
alexandair: There are two test execution paths:

- **Parallel tests** use `Invoke-ZtTestParallel`, which gets its result from `Get-ZtTestResult`. That result includes `DisplayName`.
- **Synchronous tests** use `Invoke-ZtTest`. It constructs its own result object, but does **not** add a `DisplayName` property at Invoke-ZtTest.ps1:53-71.

Synchronous tests still initially show the correct title because `Invoke-ZtTest` separately calculates `$testDisplayName` and starts the dashboard row with it at Invoke-ZtTest.ps1:77-83.

But after the test completes, the new shared finalizer calls:

```powershell
Update-ZtProgressState -WorkerName $Result.DisplayName ...
```

at Write-ZtTestFinish.ps1:60-63.

For a synchronous test, `$Result.DisplayName` is missing, so it is `$null`. `Update-ZtProgressState` replaces the existing worker record, including its `Name`. The dashboard title therefore becomes blank once that test finishes.

This affects tests deliberately kept in the main runspace—for example Data-pillar and Exchange, SharePoint, Security Compliance, and AIP tests.

src/powershell/private/tests/Write-ZtTestFinish.ps1
```diff
@@ -0,0 +1,97 @@
...
+		[AllowEmptyString()]
+		[string]
+		$LogsPath,
+
+		[switch]
+		$PassThru
+	)
+	process {
+		$Result.Messages = Get-PSFMessage -Runspace ([runspace]::DefaultRunspace.InstanceId) | Where-Object { $_ -notin $previousMessages }
+		$limit = Get-PSFConfigValue -FullName 'ZeroTrustAssessment.Tests.Statistics.MaxMessageCount'
```
alexandair: **`MaxMessageCount` does not cap retained message count**

The configuration describes a maximum number of stored messages, but Write-ZtTestFinish.ps1 retains **every** message once the threshold is exceeded; it only projects each object to fewer properties.

src/powershell/private/core/Invoke-ZtAzureRequestCache.ps1
```diff
@@ -52,12 +52,12 @@ function Invoke-ZtAzureRequestCache {
 	$cacheBlocked = Get-PSFConfigValue -FullName 'ZeroTrustAssessment.Azure.DisableCache'
 	if ($cacheBlocked) { $DisableCache = $true }
 	$isGet = -not $AzParams.ContainsKey('Method') -or $AzParams['Method'] -eq 'GET'
-	$isInCache = $script:__ZtSession.AzureCache.Value.ContainsKey($CacheKey)
+	$cachedResult = $script:__ZtSession.AzureCache.Value[$CacheKey]
 
 	# Check cache for GET requests
-	if (-not $cacheBlocked -and -not $DisableCache -and $isGet -and $isInCache) {
+	if (-not $cacheBlocked -and -not $DisableCache -and $isGet -and $cachedResult) {
```
alexandair: **Azure cache: `ContainsKey` → truthiness means empty responses are never cache hits again**

New in this PR (Invoke-ZtAzureRequestCache.ps1:55-59): a successful GET with empty/whitespace content is explicitly cached as `$null`, but the truthiness read (`… -and $cachedResult`) treats it as a miss — so such endpoints (and any `$false`/`0`/empty-array content) are re-requested on every call for the whole session. The change is *necessary* for the new expiring PSFCache (expired entries return `$null`), but if you want empty results cached, wrap them in a sentinel. Graph behavior is unchanged (old code already had an inner `if ($results)` guard).

src/powershell/private/tests/Invoke-ZtTestParallel.ps1
```diff
@@ -0,0 +1,113 @@
...
+			$result.Output = & $command @dbParam -ErrorAction Stop
+		}
+		catch {
+			throw
+		}
+
+		$result.End = Get-Date
+		$result.Duration = $result.End - $result.Start
+
+		# Reset marker in an assured way, to prevent confusion about the current test being executed
```
alexandair: **`$script:__ztCurrentTest` is not reset on failure in the parallel path**

Invoke-ZtTestParallel.ps1: the reset sits *after* a `catch { throw }`, so on failure/timeout the marker keeps pointing at the failed test until the next test overwrites it — despite the comment "Reset marker in an assured way". The sync path (Invoke-ZtTest.ps1:207-213) correctly does it in `finally`; the parallel path should too. (Incidentally, the `catch { throw }` itself is a no-op.)

src/powershell/private/tests/Start-ZtTestExecution.ps1
```diff
@@ -86,27 +89,74 @@
...
 		#endregion Calculate Resources to Import
 
+		#region Code
+		$begin = {
+			$script:ModuleRoot = $moduleRoot
+			$global:database = Connect-Database -Path $databasePath -PassThru
+			$global:msgSoFar = @{}
+		}
+		$process = {
+			if (-not $global:msgSoFar[$_.TestID]) {
```
alexandair: **Retry edge: attempt‑1 messages can be dropped from statistics**

In the worker `$process` (Start-ZtTestExecution.ps1:99-104), if the first attempt is the first test in a fresh runspace (zero prior PSF messages), `$global:msgSoFar[$_.TestID]` stays falsy, so a retry re-captures `Get-PSFMessage` — now *including* attempt‑1's messages — and excludes them from `Result.Messages`. This conflicts with the "attempts accumulate on one result object" design. Storing `@(Get-PSFMessage …)` behind a `ContainsKey` check fixes it.

src/powershell/private/tests/Start-ZtTestExecution.ps1
```diff
@@ -86,27 +89,74 @@
...
+			if ($__PSF_Agent.CurrentItem.RetryCount) { $retryCount = $__PSF_Agent.CurrentItem.RetryCount }
+			$retryTimeout = Get-PSFConfigValue -FullName 'ZeroTrustAssessment.Tests.RetryTimeout'
+			if (
+				$result.Attempts -lt $retryCount -and
+				(
+					$retryTimeout -or
+					$_ -notmatch 'Workitem timed out'
+				)
+			) { return $true }
+
```
alexandair, line 122: **Timed-out parallel tests get the generic "unexpected error" report instead of the timeout message**

In the retry condition (Start-ZtTestExecution.ps1:122-129), a non-retried timeout sets `$result.TimedOut` and then falls into `Write-ZtTestError`, which writes *"❌ Test failed due to an unexpected error"* with a raw `System.TimeoutException: Workitem timed out! …` dump (Write-ZtTestError.ps1:39-47). The sync path uses the dedicated, user-friendly message from Set-ZtTimedOutResult.ps1:24 ("did not complete within the configured timeout of…"). The dashboard also flickers: `Write-ZtTestError` sets status `Error`, then Write-ZtTestFinish.ps1:60-62 overwrites it with `TimedOut`. Since `$result.TimedOut` is computed *before* the error write, the condition can trivially branch to `Set-ZtTimedOutResult` for consistency.

alexandair (changes requested):
@FriedrichWeinmann Please, address my feedback.

src/powershell/public/Invoke-ZtAssessment.ps1
FriedrichWeinmann: `'60m'` will get you the old timeout. `'2h 30m'` for 150 minutes.
Pure numbers you already noticed resolve to seconds.
It takes Timespan string notations like `'01:00:00'` or `'1.01:30:00'`.
It also takes full Timespan objects.
And standardizes it all to Timespan.

Options I see:

- Revert to plain number and forget about it. Has the advantage of simplicity, loses the flexibility.
- Keep as is and update docs and usage. Slight breaking change (I don't expect that parameter to see too much usage, so few would be affected).
- Add a ArgumentTransformationAttribute to restore number -> Minute match in this particular case

I'm good with either choice (though admittedly, I'm a fan of the flexibility, so would prefer not to revert).
Which way do you want me to take it?

> update the parameter/docs/config contract

You mean `src/powershell/doc/readme.md:213`?

src/powershell/private/tests/Invoke-ZtTestParallel.ps1
FriedrichWeinmann: Resolved in the worker error handler, by ensuring `$result.Start` is never empty and defaults to `$result.End`.
Duration of 0 is after all correct, if the Tests never were able to start ...

src/powershell/private/tests/Invoke-ZtTest.ps1
FriedrichWeinmann: Resolved by having both Invokation commands obtain their object from the same source (`Get-ZtTestResult`).
Updated `Invoke-ZtTest` to use that same DisplayName, rather than recalculating it.
The caching nature of `Get-ZtTestResult` is of no relevant impact, as `Invoke-ZtTest` also calls `Write-ZtTestFinish` in all execution paths, which releases it.

src/powershell/private/tests/Write-ZtTestFinish.ps1
FriedrichWeinmann: While the description of the setting correctly explained its behavior, the name was indeed misleading.
Renamed to `MessageCountSoftCap`.

src/powershell/private/core/Invoke-ZtAzureRequestCache.ps1
FriedrichWeinmann: As you observed, the change was necessary (though mostly because of a concurrency issue - cached values could be expired between checking the key and retrieving the value).
I've added a mitigation - it now also checks whether it contains the key ... but only after already retrieving the value and only if the value retrieved resolves to $false.

src/powershell/private/tests/Invoke-ZtTestParallel.ps1
FriedrichWeinmann: Excellent point, added a finally and moved it there.

src/powershell/private/tests/Start-ZtTestExecution.ps1
FriedrichWeinmann: Interesting observation. Adjusted as suggested.

src/powershell/private/tests/Start-ZtTestExecution.ps1
FriedrichWeinmann: Excellent point, added that as suggested.

src/powershell/public/Invoke-ZtAssessment.ps1
astaykov: Agreed upon:

* Add a ArgumentTransformationAttribute to restore number -> Minute match in this particular case

@FriedrichWeinmann can you update this.

FriedrichWeinmann: Implemented

PR #1319, opened 2026-06-17
Add Synchronization.Read.All scope for provisioning test (21886)

alflokken:
### Summary 

Test 21886 ("Applications are configured for automatic user provisioning") queries the following synchronization endpoints for each SSO-enabled service principal:

- servicePrincipals/{id}/synchronization/templates
- servicePrincipals/{id}/synchronization/jobs

These endpoints require the `Synchronization.Read.All` Graph scope, which was missing from the assessment's requested permissions. As a result, the test returns 401 Unauthorized under the default consent set and cannot run successfully out of the box.

### Fix

- Added `Synchronization.Read.All` to `Get-ZtGraphScope.ps1`.
- Added the scope to the consent list in `src/powershell/doc/readme.md`.

### Impact

- Adds one read-only delegated scope.
- Existing users will see a one-time re-consent prompt the next time they connect.
- No code or output changes. This is a permissions-only fix that allows test 21886 to execute successfully.

### Testing

- Verified the two endpoints used by 21886 require `Synchronization.Read.All`.
- Confirmed the scope is now requested at connect time and the test no longer returns 401.


astaykov:
Hi @alflokken

Thanks for reporting this. After analysis this is the root cause for test 21886 to error:
>The API used (https://learn.microsoft.com/en-us/graph/api/synchronization-synchronization-list-templates?view=graph-rest-1.0&tabs=http) indeed manifests requirement for `Synchronization.Read.All` permission. However, reading out synchronization jobs is also possible with `Application.Read.All` (already requested) and if the calling user holds an active privileged role, like Cloud Application Administrator. Indeed, if the calling user identity only has `Global Reader` role, the synchronization API will return 401 with unknown error body.

For the pull request - please do not create pull requests without first creating issue, describing what you observe and how you use the product. And, even more important - do not edit the metadata that is not related to the issue you are trying to fix. I will close this PR and create an issue for tracking and implementing. Until we add this permission, you can sign-in with more privileged role, like Cloud Application Administrator and 21886 test will successfully execute.

Issue #1330, opened 2026-06-22
Module Sharepoint Not working in Powershell 7 (0x80131515)


rjimenezm67:
Hi,
I am running ZeroTrustAssessment in PowerShell 7 only, as documented by Microsoft. The issue occurs during the SharePoint Online connection phase.
The problem is not that I am using Windows PowerShell 5.1 interactively. I am not.
The failure occurs when PowerShell 7 attempts to import the SharePoint Online module using the Windows PowerShell compatibility layer (-UseWindowsPowerShell).
Microsoft documentation states:

ZeroTrustAssessment should be run from a PowerShell 7 window. [learn.microsoft.com], [github.com]
In a PowerShell 7 console, Microsoft.Online.SharePoint.PowerShell must be imported with -UseWindowsPowerShell. 
(https://learn.microsoft.com/en-us/security/zero-trust/assessment/get-started), [learn.microsoft.com]
PowerShell 7 compatibility mode creates a background Windows PowerShell 5.1 process/session (WinPSCompatSession). (https://learn.microsoft.com/en-us/powershell/sharepoint/sharepoint-online/connect-sharepoint-online), [[learn.microsoft.com]](https://learn.microsoft.com/es-es/powershell/sharepoint/sharepoint-online/connect-sharepoint-online)

Environment

Interactive shell: PowerShell 7
ZeroTrustAssessment: latest preview / 2.4.x
SharePoint module:

Microsoft.Online.SharePoint.PowerShell
Version: 16.0.27313.12000

D:\_Scripts\Modules\Microsoft.Online.SharePoint.PowerShell\16.0.27313.12000

Import-Module "D:\_Scripts\Modules\Microsoft.Online.SharePoint.PowerShell\16.0.27313.12000\Microsoft.Online.SharePoint.PowerShell.psd1" -UseWindowsPowerShell -Verbose
Failure from remote command:
Could not load file or assembly 'file:///D:\_Scripts\Modules\Microsoft.Online.SharePoint.PowerShell\16.0.27313.12000\Microsoft.Online.SharePoint.PowerShell.dll'
Operation is not supported. (Exception from HRESULT: 0x80131515)

Impact
Connect-ZtAssessment is unable to connect to SharePoint Online, and SharePoint-related tests are skipped.

Question
Is this a known limitation of:

PowerShell 7 compatibility mode (WinPSCompatSession) for this module, or
the current SharePoint Online module loading path used by Connect-ZtAssessment?

[...]

PR #1335, opened 2026-06-22
Mark of the Web unblock when loading dependencies

rberghuis:
0x80131515 is the result of Windows Mark of the Web (Zone.Identifier) block. During dependency loading, confirm if MotW would block load and unblock if applicable. Resolves issue #1330


astaykov:
Hi, @rberghuis 

Instead of doing PR, could you, please, first try opening an issue, describing your exact problem, setup, environment and steps to reproduce.

Thank you

PR #1336, opened 2026-06-22
Empty array handling for ReportTemplate.html

rberghuis:
The Mobile Devices diagram received data but did no validation before handing it to the library. Two things were missing:

**1. It passed zero-value links**
If a tenant has no mobile devices, all the mobile links have value 0. Those zeros were passed straight to d3-sankey, which tried to draw a chart with no meaningful data.

**2. It had no guard for empty data.**
When all links are zero, filtering them out leaves an empty dataset. d3-sankey then tried to compute `new Array(max_depth + 1)` where `max_depth` is `undefined` (no nodes to measure), which evaluates to `new Array(NaN)`, an invalid operation that crashes the whole page.

**The fix adds two lines of defence:**
1. Strip out any links with a value of zero or null before doing anything else.
2. If nothing is left after stripping, show "No data available." instead of attempting to render the chart.


astaykov:
This issue is fixed starting version 2.4.13-preview and later - use the lastest preview version to validate: https://www.powershellgallery.com/packages/ZeroTrustAssessment/2.4.22-preview.

For the future, please do not raise a PR without associated issue. If you find an issue, please report.

P.S. The template is not a static file. It is being generated dynamically (react app) if changes are required.

PR #1337, opened 2026-06-22
fix: update KnownPillars parameter to use 'security-ops' instead of '…

mzmaili:
…secops'


astaykov (approved):
This was my over-optimization enthusiasm!

Thanks for fixing!

PR #1339, opened 2026-06-23
Devices - 51021 - Least privilege Intune RBAC roles are assigned instead of relying only on full Intune Administrator

praneeth-0000:


astaykov:
Spec needs rewriting. Some APIs are missing. This check cannot be automated in a reliable way with the existing APIs.

PR #1340, opened 2026-06-23
add SecurityIdentitiesSensors.Read.All in scope for secops pillar

praneeth-0000:
The permission is needed for 41002, 41003 SecOps Tests


alexandair (approved):
LGTM

PR #1341, opened 2026-06-23
Device - 51001 - New Spec changes 

Manoj-Kesana:


src/powershell/tests/Test-Assessment.51001.ps1
```diff
@@ -27,6 +27,24 @@ function Test-Assessment-51001 {
...
+        }).Count -gt 0
+    }
+    catch {
+        $licenseQueryFailed = $true
+        Write-PSFMessage "Failed to retrieve subscribed SKUs: $_" -Tag Test -Level Warning
+    }
+
+    # Q2: Retrieve all EPM elevation settings and elevation rules policies
     $sql = @"
     SELECT id, name, platforms, technologies, isAssigned, templateReference, to_json(assignments) as assignments
```
alexandair, line 51: **Dead code + inconsistent assignment determination**

The Data Collection region selects and parses `assignments`:
```powershell
SELECT ... to_json(assignments) as assignments ...
foreach ($policy in $epmPolicies) {
    if ($policy.assignments -is [string]) {
        $policy.assignments = $policy.assignments | ConvertFrom-Json
    }
    ...
}
```
…but assignment status is then decided purely from `$_.isAssigned`. The parsed `assignments` is **never read** — dead code. The established sibling pattern (e.g. `Test-Assessment.24540`) determines assignment via `$_.assignments -and $_.assignments.Count -gt 0`. Recommend either:
- switch the check to `assignments.Count -gt 0` (consistent with sibling Intune/ConfigurationPolicy tests, removes the dead parse), **or**
- if `isAssigned` is the intentional source of truth, drop the `to_json(assignments)` select and the parsing loop entirely.

src/powershell/tests/Test-Assessment.51001.ps1
```diff
@@ -42,42 +60,46 @@ function Test-Assessment-51001 {
...
+    $rulesPolicies    = @($epmPolicies | Where-Object { $_.PolicyType -eq 'Elevation Rules' })
 
     $settingsAssigned = @($settingsPolicies | Where-Object { $_.isAssigned }).Count -gt 0
     $rulesAssigned    = @($rulesPolicies    | Where-Object { $_.isAssigned }).Count -gt 0
     #endregion Data Collection
 
     #region Assessment Logic
-    if ($epmPolicies.Count -eq 0) {
-        Add-ZtTestResultDetail -SkippedBecause NotApplicable
+    if ($licenseQueryFailed) {
```
alexandair, line 81: **Investigate message mismatch**

Only the **license** Graph call (`subscribedSkus`) is wrapped in try/catch, but the Investigate message describes the *policies* API:
> "The Intune Endpoint Privilege Management **policies API** returned an authorization (401/403)…"

The policies actually come from the local DuckDB table — which can't return 401/403. So the message describes a failure mode that can't trigger here, and the real trigger (a failed `subscribedSkus`/licensing call) is mislabeled. Recommend rewording the Investigate text to reference the **licensing / subscribedSkus** call.

alexandair (changes requested):

alexandair (approved):
LGTM

PR #1345, opened 2026-06-24
Adding alias `Tenant` to the `TenantId` parameter

astaykov:
The parameter works equally same with both the tenant GUID and any verified domain name. The name TenantId very binding to the Guid. Adding `Tenant` as alias also aligns to [Connect-AzAccount](https://learn.microsoft.com/en-us/powershell/module/az.accounts/connect-azaccount?view=azps-16.0.0)'s `Tenant` property. This addition does not functionally change anything.

### DIFF
--- a/src/powershell/public/Connect-ZtAssessment.ps1
+++ b/src/powershell/public/Connect-ZtAssessment.ps1
@@ -20,7 +20,7 @@ function Connect-ZtAssessment {
 		Uses Graph Powershell's cached authentication tokens.
 
 	.PARAMETER TenantId
-		The tenant ID to connect to. If not specified, the default tenant will be used.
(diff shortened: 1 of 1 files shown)


src/powershell/public/Connect-ZtAssessment.ps1
```diff
@@ -72,6 +72,7 @@ function Connect-ZtAssessment {
 		$UseTokenCache,
 
 		[string]
+		[Alias('Tenant')]
```
alexandair, line 74: Common way of specifying the alias attribute is:
```powershell
[Alias('Tenant')]
[string]
$TenantId,
```

alexandair (changes requested):

alexandair (approved):
LGTM

PR #1357, opened 2026-06-26
Add SecurityAlert.Read.All scope

praneeth-0000:
Add SecurityAlert.Read.All scope to documentation and Get-ZtGraphScope function. Required by 41019


astaykov (approved):
LGTM

Issue #1360, opened 2026-06-29
Tests that require EXO, SPO or Sercurity & Compliance PowerShell cannot run in parallel


astaykov:
This issue is to keep track of infrastructure change:
Tests that use Exchange Online, SharePoint Online, or Security and Compliance PowerShell modules cannot be run in parallel, because the respective modules are not loaded into the run spaces spawn for parallel test execution.
Currently there is a partial implementation that forces all Data pillar tests and all tests requiring Security and Compliance PowerShell to be run on the main thread.

PR #1361, opened 2026-06-29
Tests requiring EXO or SPO must not be run in parallel. 

astaykov:
The current check only does this for _all_ Data pillar tests, or tests that use SecurityCompliance service. This change adds EXO and SPO to the SecurityCompliance as services.

Potential further improvement - completely remove the Pillar check and only leave the required services check. This is more reliable and even can have the side effect of pushing Data checks to parallel if there is now (or in the future) such that only use Graph or Azure.

Addresses #1360

### DIFF
--- a/src/powershell/private/tests/Invoke-ZtTests.ps1
+++ b/src/powershell/private/tests/Invoke-ZtTests.ps1
@@ -118,9 +118,9 @@
 	$testsToRun = $testsToRun.Where{ $_.TestId -notin $skippedTestsForService.TestId }
 
 	# Separate Sync Tests (Compliance/ExchangeOnline/SharePointOnline/AIP) from Parallel Tests (because of DLL order to manage in runspaces & remoting into WPS)
-	# Tests that depend on SecurityCompliance/ExchangeOnline/SharePointOnline must run on the main thread
+	# Tests that depend on SecurityCompliance/ExchangeOnline/SharePointOnline/AIPService must run on the main thread
 	# regardless of pillar: those services expose their cmdlets through dynamically-generated, connection-bound
-	# proxy modules (Connect-IPPSSession / Connect-ExchangeOnline / Connect-SPOService) that only exist in the
+	# proxy modules (Connect-IPPSSession / Connect-ExchangeOnline / Connect-SPOService / Connect-AipService) that only exist in the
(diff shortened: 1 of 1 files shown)


src/powershell/private/tests/Invoke-ZtTests.ps1
```diff
@@ -111,15 +111,21 @@
...
+	# proxy modules (Connect-IPPSSession / Connect-ExchangeOnline / Connect-SPOService) that only exist in the
+	# main runspace where Connect-ZtAssessment ran. In a worker runspace those cmdlets are "not recognized",
+	# so such tests would incorrectly skip (e.g. Get-SafeLinksPolicy, Get-OrganizationConfig, Get-SPOTenant).
+	[string[]]$mainThreadServices = @('SecurityCompliance', 'ExchangeOnline', 'SharePointOnline')
 	[int[]]$syncTestIds   = $testsToRun.Where{
-		$_.Pillar -contains 'Data' -or $_.Service -contains 'SecurityCompliance'
+		$_.Pillar -contains 'Data' -or ($_.Service | Where-Object { $_ -in $mainThreadServices })
 	}.TestId
 	$syncTests     = $testsToRun.Where{ $_.TestId -in $syncTestIds }
 	$parallelTests = $testsToRun.Where{ $_.TestId -notin $syncTestIds }
```
alexandair:
```powershell
$mainThreadServices = 'SecurityCompliance', 'ExchangeOnline', 'SharePointOnline'

$syncTests, $parallelTests = $testsToRun.Where({
    $_.Pillar -contains 'Data' -or
    @($_.Service).Where({ $_ -in $mainThreadServices })
}, 'Split')
```

Is `AipService` skipped on purpose?

### What improved

**One pass instead of three.** The original computed `$syncTestIds`, then re-scanned for `$syncTests`, then re-scanned again for `$parallelTests`.
**No intermediate ID round-trip.** We were converting objects → TestIds → back to objects via -in. Split keeps the objects directly, so it's also impossible for the two sets to drift out of sync.
**No duplicated/negated predicate.** Split guarantees parallelTests is the exact complement; you don't maintain a -notin mirror.
`@($_.Service).Where({...})` instead of `$_.Service | Where-Object {...}` avoids the pipeline overhead, and `@()` still handles the single-vs-multi Service case.

alexandair (changes requested):

alexandair (approved):
LGTM

Issue #1365, opened 2026-06-30
Build infrastructure and CI failing in newest PowerShell and clean state VMs


astaykov:
Couple of infrastructure components would fail on newer PowerShell versions (7.6.3+) and clean images. This is related to how PS registers PSGallery as valid source and general push to move away from Install-Module towards PSResourceGet.
Issue becoming visible in our CI builds, e.g. https://github.com/microsoft/zerotrustassessment/actions/runs/28400507463/job/84150219451

PR #1367, opened 2026-06-30
#1365 fix PSGallery registration

astaykov:
#1365 fix PSGallery reigstration for build infrastructure components


alexandair (approved):
LGTM

PR #1369, opened 2026-06-30
SecOps - 41029 - Anti-malware policies are configured according to recommended settings

Manoj-Kesana:


src/powershell/tests/Test-Assessment.41029.ps1
```diff
@@ -0,0 +1,193 @@
...
+            }
+            else {
+                $passed = $true
+                $testResultMarkdown = "✅ All in-scope anti-malware policies have the Common Attachment Filter enabled, zero-hour auto purge enabled, and route detected malware to an admin-only quarantine."
+            }
+        }
+    }
+    #endregion Assessment Logic
+
+    #region Report Generation
```
alexandair, line 165: **Missing informational summary line (spec requirement, silently dropped)**

The spec's "Test output data" section explicitly requires, below the per-policy table:

> Of N in-scope policies, X have `EnableInternalSenderAdminNotifications == True` and Y have `EnableExternalSenderAdminNotifications == True`. Microsoft's baseline recommends both off; these settings do not affect this check's Pass/Fail.

The implementation collects `EnableInternalSenderAdminNotifications` and `EnableExternalSenderAdminNotifications` in Data Collection (line ~40) but **never uses them anywhere** — no counting logic, no summary line rendered in Report Generation. This is a real data point the spec calls for and it's dropped entirely. Add the counts and summary line after the per-policy table (or truncation notice) in the `#region Report Generation` block, e.g.:

```powershell
$internalNotifyCount = @($inScopePolicies | Where-Object { $_.EnableInternalSenderAdminNotifications -eq $true }).Count
$externalNotifyCount = @($inScopePolicies | Where-Object { $_.EnableExternalSenderAdminNotifications -eq $true }).Count
$testResultMarkdown += "`n_Of $totalInScope in-scope policies, $internalNotifyCount have internal sender admin notifications enabled and $externalNotifyCount have external sender admin notifications enabled. Microsoft's baseline recommends both off; these settings do not affect this check's Pass/Fail._`n"
```

src/powershell/tests/Test-Assessment.41029.ps1
```diff
@@ -0,0 +1,193 @@
...
+    # Step 2: Empty policy result indicates API failure — default policy always exists in Exchange Online
+    if ($policies.Count -eq 0) {
+        $customStatus = 'Investigate'
+        $testResultMarkdown = "⚠️ **Investigate:** ``Get-MalwareFilterPolicy`` returned no results. The default policy is always present in Exchange Online; an empty result indicates an API or connectivity failure."
+    }
+    else {
+        # Step 3: Build in-scope set = default policy + policies referenced by enabled rules
+        $enabledRules = @($rules | Where-Object { $_.State -eq 'Enabled' })
+
+        # Build case-insensitive lookup by Identity for rule-to-policy resolution
```
alexandair: **Rule-to-policy matching uses `Identity` only, spec says "Identity or Name".**

 Step 4 of the spec reads: *"…does not match any policy `Identity` or `Name`…"*. The code's `$policyByIdentity` lookup is keyed solely on `policy.Identity`, and the `Select-Object` in Data Collection doesn't even retrieve a separate `Name` property. In practice, for `Get-MalwareFilterPolicy` the `Identity` returned by the cmdlet *is* the policy name (per Microsoft docs, Identity accepts "name, GUID, or DN" as inputs, and the returned `Identity` is the name string), so this is unlikely to cause false orphan detections in real tenants. Still, worth a quick sanity check/comment acknowledging the spec's dual-key language, since the current implementation only defends against one of the two identifiers the spec calls out.

src/powershell/tests/Test-Assessment.41029.ps1
```diff
@@ -0,0 +1,193 @@
...
+    #endregion Assessment Logic
+
+    #region Report Generation
+    if ($inScopePolicies.Count -gt 0) {
+        $totalInScope = $inScopePolicies.Count
+        $displayPolicies = $inScopePolicies | Select-Object -First 10
+        $truncated = $totalInScope -gt 10
+
+        $testResultMarkdown += "`n`n### [In-scope anti-malware policies](https://security.microsoft.com/antimalwarev2)`n`n"
+        $testResultMarkdown += "| Identity | Is Default | Common Attachment Filter | File Types (count) | ZAP | Quarantine Tag | Applied | Result |`n"
```
alexandair: I think we agreed to capitalize only the first word for column names.

src/powershell/tests/Test-Assessment.41029.ps1
```diff
@@ -0,0 +1,193 @@
...
+
+    $passed = $false
+    $customStatus = $null
+    $inScopePolicies = [System.Collections.Generic.List[object]]::new()
+    $orphanedRules = @()
+
+    # Step 2: Empty policy result indicates API failure — default policy always exists in Exchange Online
+    if ($policies.Count -eq 0) {
+        $customStatus = 'Investigate'
+        $testResultMarkdown = "⚠️ **Investigate:** ``Get-MalwareFilterPolicy`` returned no results. The default policy is always present in Exchange Online; an empty result indicates an API or connectivity failure."
```
alexandair, line 80: Our convention is to use `%TestResult%` in `$testResultMarkdown` and later:
```
 $testResultMarkdown = $testResultMarkdown -replace '%TestResult%', $mdTable
```

Please, refactor it.

src/powershell/tests/Test-Assessment.41029.ps1
```diff
@@ -0,0 +1,193 @@
...
+        foreach ($policy in $displayPolicies) {
+            $identity       = Get-SafeMarkdown -Text $policy.Identity
+            $isDefault      = if ($policy.IsDefault -eq $true) { 'Yes' } else { 'No' }
+            $fileFilter     = if ($policy.EnableFileFilter -eq $true) { '✅ Yes' } else { '❌ No' }
+            $fileTypesCount = if ($policy.FileTypes) { $policy.FileTypes.Count } else { 0 }
+            $zap            = if ($policy.ZapEnabled -eq $true) { '✅ Yes' } else { '❌ No' }
+            $quarantineTag  = Get-SafeMarkdown -Text $policy.QuarantineTag
+            $applied        = if ($policy.IsDefault -eq $true) { 'Default catch-all' } else { 'Via enabled rule' }
+
+            # Per-row result is independent of the overall $customStatus
```
alexandair: The row-level `$rowResult` computation in Report Generation re-derives the same `QuarantineTag`/`EnableFileFilter`/`ZapEnabled` logic already computed in the Assessment Logic loop (`$investigatePolicies`/`$nonCompliantPolicies`). Consider tagging each in-scope policy with its computed status once (e.g., add a `RowResult` note property when building `$inScopePolicies`) to avoid maintaining the compliance rule in two places — low risk today since both copies are correct, but a future rule change would need to be updated in two spots.

alexandair (changes requested):

alexandair (approved):
LGTM

PR #1372, opened 2026-07-01
SecOps - 41200 - Microsoft Sentinel is enabled on every Log Analytics workspace used for security monitoring

aahmed-spec:
Added SecOps Pillar


alexandair (approved):
LGTM

PR #1384, opened 2026-07-02
Group Membership Resolution - Increased pagesize

FriedrichWeinmann:
Resolving members for large groups does not, after all, have to take half an hour.
Updated page-size from default to maximum


astaykov (approved):
LGTM

PR #1385, opened 2026-07-03
SecOps 41053 - Network protection is enabled in block mode

ashwinikarke:


src/powershell/tests/Test-Assessment.41053.ps1
```diff
@@ -0,0 +1,136 @@
...
+    Workshop Task: SECOPS-053
+    Pillar: SecOps
+    Category: Endpoint threat protection
+    Required permission: SecurityEvents.Read.All
+#>
+
+function Test-Assessment-41053 {
+    [ZtTest(
+        Category = 'Endpoint threat protection',
+        CompatibleLicense = ('WINDEFATP'),
```
alexandair: **License value inconsistency with a near-identical sibling test (41060).**

`ATA` was team-confirmed (@astaykov knows more about it) as the correct `CompatibleLicense` for MDATP/Defender-for-Endpoint Secure Score checks (used in Test-Assessment.41060.ps1, which follows the exact same `secureScoreControlProfiles`/`MDATP` pattern). `WINDEFATP` appears nowhere else in the codebase. The code correctly matches its own spec, but this may indicate the **spec** itself has a license-value error worth double-checking with the author, since 41053 and 41060 are both MDATP-based Secure Score checks that would normally share the same license gate.

alexandair (changes requested):

alexandair (approved):
LGTM

PR #1386, opened 2026-07-06
Implement 41032 (https://github.com/microsoft/ztspecs/issues/857)

astaykov:
Note: intentionally not included Graph queries: as per spec they are optional and do not influence the result of the check. Only bring additional information. Given the current performance issues, decided to not implement optional queries.

### DIFF
--- a/src/powershell/tests/Test-Assessment.41032.ps1
+++ b/src/powershell/tests/Test-Assessment.41032.ps1
@@ -40,7 +40,6 @@ function Test-Assessment-41032 {
     $params = @{
         TestId = '41032'
         Title  = 'Safe Links policies in Microsoft Defender for Office 365 are configured to scan and block malicious URLs'
-        Status = $false
     }
 
     #region Data Collection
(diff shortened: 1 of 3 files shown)


src/powershell/tests/Test-Assessment.41032.ps1
```diff
@@ -0,0 +1,270 @@
...
+    )]
+    [CmdletBinding()]
+    param()
+
+    # Invariant result metadata — TestId and Title are defined once here. Each return path only sets
+    # the varying fields (Status, Result, and optionally CustomStatus) before Add-ZtTestResultDetail.
+    $params = @{
+        TestId = '41032'
+        Title  = 'Safe Links policies in Microsoft Defender for Office 365 are configured to scan and block malicious URLs'
+        Status = $false
```
alexandair: All four `Add-ZtTestResultDetail` calls explicitly set `$params.Status` before use, so the initial `Status = $false` in this hash table is redundant.

alexandair (changes requested):

alexandair (approved):
LGTM

PR #1389, opened 2026-07-08
SecOps - 41059 - EDR in block mode is enabled when Microsoft Defender Antivirus is not the primary antivirus

sandeepjha000:
[Spec-41059](https://github.com/microsoft/ztspecs/blob/main/specs/secops/41059.md)

[Issue #908](https://github.com/microsoft/ztspecs/issues/908)

### DIFF
--- a/src/powershell/tests/Test-Assessment.41059.ps1
+++ b/src/powershell/tests/Test-Assessment.41059.ps1
@@ -69,6 +69,7 @@ function Test-Assessment-41059 {
 
     # Q2 – Retrieve the most recent Secure Score snapshot (only if Q1 succeeded).
     $latestSecureScore = $null
+    $errorMsgQ2        = $null
 
     if ($null -ne $controlProfile) {
         Write-ZtProgress -Activity $activity -Status 'Retrieving latest Microsoft Secure Score'
@@ -77,7 +78,8 @@ function Test-Assessment-41059 {
             $latestSecureScore = $scoreResponse.value | Select-Object -First 1
         }
         catch {
-            Write-PSFMessage "Failed to retrieve Secure Score snapshot: $_" -Level Warning
+            $errorMsgQ2 = $_
+            Write-PSFMessage "Failed to retrieve Secure Score snapshot: $errorMsgQ2" -Level Warning
         }
     }
 
@@ -117,9 +119,15 @@ function Test-Assessment-41059 {
     $profileTitle = $controlProfile.title
     $maxScore     = $controlProfile.maxScore
 
-    # ── Investigate: Q2 returned no snapshot ──
-    if ($null -eq $latestSecureScore) {
-        $testResultMarkdown = '⚠️ The EDR in block mode control profile exists but the current Microsoft Secure Score snapshot could not be retrieved.'
+    # ── Investigate: Q2 failed or returned no snapshot ──
+    if ($null -ne $errorMsgQ2 -or $null -eq $latestSecureScore) {
+        if ($null -ne $errorMsgQ2) {
+            $investigateReason = 'Microsoft Graph returned an unexpected error retrieving the latest Secure Score snapshot. Re-run the assessment in 5–10 minutes.'
+        }
+        else {
+            $investigateReason = 'The EDR in block mode Secure Score control or latest Secure Score snapshot could not be located.'
+        }
+        $testResultMarkdown = "⚠️ $investigateReason"
         $customStatus       = 'Investigate'
 
         $params = @{
--- a/src/powershell/tests/Test-Assessment.41059.ps1
+++ b/src/powershell/tests/Test-Assessment.41059.ps1
@@ -191,7 +191,7 @@ function Test-Assessment-41059 {
     $mdTable = @"
 
 $mdFailLink
-| Control title | Control id | Score | Max Score | Ignored | Last modified | Status |
+| Control title | Control id | Score | Max score | Ignored | Last modified | Status |
 | :------------ | :--------- | :---- | :-------- | :------ | :------------ | :----- |
 $tableRows
 "@
--- a/src/powershell/tests/Test-Assessment.41059.ps1
+++ b/src/powershell/tests/Test-Assessment.41059.ps1
@@ -117,7 +117,6 @@ function Test-Assessment-41059 {
     }
 
     $profileTitle = $controlProfile.title
-    $maxScore     = $controlProfile.maxScore
 
     # ── Investigate: Q2 failed or returned no snapshot ──
     if ($null -ne $errorMsgQ2 -or $null -eq $latestSecureScore) {
@@ -149,9 +148,13 @@ function Test-Assessment-41059 {
             Select-Object -First 1
     }
 
-    # ── Investigate: profile exists but snapshot has no scored entry for this control ──
-    if ($null -eq $controlScoreEntry) {
-        $testResultMarkdown = "⚠️ The EDR in block mode control profile (**$controlId**) exists but the latest Secure Score snapshot has no scored entry for this control."
+    # ── Investigate: profile exists but snapshot has no scored entry for this control,
+    #    or score/maxScore is null (would produce a false Pass via PowerShell $null coercion) ──
+    $score    = $controlScoreEntry.score
+    $maxScore = $controlProfile.maxScore
+
+    if ($null -eq $controlScoreEntry -or $null -eq $score -or $null -eq $maxScore) {
+        $testResultMarkdown = '⚠️ The EDR in block mode Secure Score control or latest Secure Score snapshot could not be located.'
         $customStatus       = 'Investigate'
 
         $params = @{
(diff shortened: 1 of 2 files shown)


src/powershell/tests/Test-Assessment.41059.ps1
```diff
@@ -0,0 +1,214 @@
...
+        $params = @{
+            TestId       = '41059'
+            Title        = 'EDR in block mode is enabled when Microsoft Defender Antivirus is not the primary antivirus'
+            Status       = $passed
+            Result       = $testResultMarkdown
+            CustomStatus = $customStatus
+        }
+        Add-ZtTestResultDetail @params
+        return
+    }
```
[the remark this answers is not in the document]
sandeepjha000: Test should not be skipped (code is correct and setting to investigate as confirmed by spec author on issue https://github.com/microsoft/ztspecs/issues/908)
Updated .DESCRIPTION comments to mention investigate status for 401/403 errors, similar to test 41009 - [eddf32a](https://github.com/microsoft/zerotrustassessment/pull/1389/commits/eddf32a7de612398686b7a1456fc9719533466d7)

src/powershell/tests/Test-Assessment.41059.ps1
```diff
@@ -0,0 +1,214 @@
...
+    if ($null -ne $controlProfile) {
+        Write-ZtProgress -Activity $activity -Status 'Retrieving latest Microsoft Secure Score'
+        try {
+            $scoreResponse     = Invoke-ZtGraphRequest -RelativeUri 'security/secureScores' -Top 1 -ApiVersion beta -DisablePaging -ErrorAction Stop
+            $latestSecureScore = $scoreResponse.value | Select-Object -First 1
+        }
+        catch {
+            Write-PSFMessage "Failed to retrieve Secure Score snapshot: $_" -Level Warning
+        }
+    }
```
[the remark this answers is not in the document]
sandeepjha000: Both endpoints use the same SecurityEvents.Read.All permission and the same auth token. If Q1 succeeds, a 401/403 on Q2 is practically impossible.
Added condition to handle unexpected API error for Q2: [f222a99](https://github.com/microsoft/zerotrustassessment/pull/1389/commits/f222a9920ce6d13d74dde53ad29973c5111cdf8b)

src/powershell/tests/Test-Assessment.41059.ps1
```diff
@@ -0,0 +1,214 @@
...
+        $params = @{
+            TestId       = '41059'
+            Title        = 'EDR in block mode is enabled when Microsoft Defender Antivirus is not the primary antivirus'
+            Status       = $passed
+            Result       = $testResultMarkdown
+            CustomStatus = $customStatus
+        }
+        Add-ZtTestResultDetail @params
+        return
+    }
```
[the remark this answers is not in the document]
sandeepjha000: Not Applicable - as mentioned in prior comment, if there is a 401/403 error, we set status to investigate rather than skipping the test in Q1. So the test would never reach to Q2 in case of permissions issue.

src/powershell/tests/Test-Assessment.41059.ps1
```diff
@@ -0,0 +1,214 @@
...
+
+    # ── Determine ignored state from the most recent controlStateUpdates entry ──
+    $latestStateUpdate = @($controlProfile.controlStateUpdates | Sort-Object { if ($_.updatedDateTime) { [datetime]$_.updatedDateTime } else { [datetime]::MinValue } } -Descending) | Select-Object -First 1
+    $isIgnored         = $latestStateUpdate -and $latestStateUpdate.state -eq 'Ignored'
```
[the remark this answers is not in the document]
sandeepjha000: Not applicable - PowerShell's -eq is already case-insensitive by default. The -ieq change is redundant.

src/powershell/tests/Test-Assessment.41059.ps1
```diff
@@ -0,0 +1,222 @@
...
+    #endregion Data Collection
+
+    #region Assessment Logic
+
+    $passed       = $false
+    $customStatus = $null
+
+    # ── Q1 failed: 401/403, unexpected error, or missing profile → Investigate ──
+    if ($null -eq $controlProfile) {
+        if ($httpStatusQ1 -in @(401, 403)) {
```
alexandair, line 95: **401/403 results are incorrectly reported as Investigate instead of Skipped.**  

The specification explicitly requires **Skipped** when either Graph query returns HTTP 401 or 403. The Q1 branch assigns `CustomStatus = 'Investigate'` instead, and the Q2 catch does not retain or evaluate an HTTP status at all. This misclassifies a permissions/connectivity prerequisite as an assessment result. It also contradicts the file’s own synopsis. Handle 401/403 from **both** requests with `Add-ZtTestResultDetail -SkippedBecause ...` followed by `return`.

src/powershell/tests/Test-Assessment.41059.ps1
```diff
@@ -0,0 +1,222 @@
...
+            CustomStatus = $customStatus
+        }
+        Add-ZtTestResultDetail @params
+        return
+    }
+
+    # ── Determine ignored state from the most recent controlStateUpdates entry ──
+    $latestStateUpdate = @($controlProfile.controlStateUpdates | Sort-Object { if ($_.updatedDateTime) { [datetime]$_.updatedDateTime } else { [datetime]::MinValue } } -Descending) | Select-Object -First 1
+    $isIgnored         = $latestStateUpdate -and $latestStateUpdate.state -eq 'Ignored'
+
```
alexandair, line 174: **Nullable score fields can produce a false Pass.**  

Graph marks both `controlScore.score` and `secureScoreControlProfile.maxScore` as nullable. In PowerShell, `$null -ge $null` and a numeric value `-ge $null` evaluate to `$true`. Therefore a returned score entry with missing `score` and/or `maxScore` can pass the test. Require both values to be non-null before comparison; absent scoring data should return `Investigate`.

src/powershell/tests/Test-Assessment.41059.ps1
```diff
@@ -0,0 +1,222 @@
...
+        $profileResults = Invoke-ZtGraphRequest -RelativeUri 'security/secureScoreControlProfiles' -Filter "service eq 'MDATP' and id eq '$controlId'" -ApiVersion beta -ErrorAction Stop
+        $controlProfile = $profileResults | Select-Object -First 1
+    }
+    catch {
+        $errorMsgQ1   = $_
+        $httpStatusQ1 = Get-ZtHttpStatusCode -ErrorRecord $_
+        Write-PSFMessage "Failed to retrieve EDR in block mode control profile: $errorMsgQ1" -Level Warning
+    }
+
+    # Q2 – Retrieve the most recent Secure Score snapshot (only if Q1 succeeded).
```
alexandair, line 70: **The second Graph request is not made when Q1 returns no profile or errors.**

The specification says to **call both GET requests**. The short-circuit does not change the final ambiguous result, but it means missing snapshot data cannot be distinguished and does not precisely implement the requested query flow.

Is it discussed with the spec author to run Q2 only when Q1 succeed?

alexandair (changes requested):

src/powershell/tests/Test-Assessment.41059.ps1
sandeepjha000: Not Applicable - Spec author confirmed over the issue to set status to Investigate for 401/403 error. 
If Q1 fails with 401/403, Q2 will also fail as they both use same permissions ( _SecurityEvents.Read.All_ ).

src/powershell/tests/Test-Assessment.41059.ps1
sandeepjha000: when Q1 returns no profile, Q2's snapshot is of no use,  you can't look up scid_2004 in controlScores without a confirmed profile. The short-circuit is a practical optimisation with no impact on the result

src/powershell/tests/Test-Assessment.41059.ps1
alexandair: Shouldn't they update the spec after that?
Isn't the point of the spec issue to correct the spec?

src/powershell/tests/Test-Assessment.41059.ps1
alexandair: Again, the spec should be updated.
I know that we don't sync the spec and test later after some bug fixes, but the spec should be in good shape for at least the first review before it's merged, don't you think?

src/powershell/tests/Test-Assessment.41059.ps1
sandeepjha000: Agree, I have requested to update the spec

src/powershell/tests/Test-Assessment.41059.ps1
sandeepjha000: Yes, Agree. I have requested to update the spec

src/powershell/tests/Test-Assessment.41059.ps1
sandeepjha000: fixed in commit: [7598f8b](https://github.com/microsoft/zerotrustassessment/pull/1389/commits/7598f8b740f2234fbb34671729afb725e725f7a4)

src/powershell/tests/Test-Assessment.41059.ps1
sandeepjha000: spec updated

src/powershell/tests/Test-Assessment.41059.ps1
sandeepjha000: spec updated

alexandair (approved):
LGTM

PR #1397, opened 2026-07-13
SecOps - 41033 - Anti-phishing policies in Microsoft Defender for Office 365 are configured with impersonation and spoof protection

Manoj-Kesana:


src/powershell/tests/Test-Assessment.41033.ps1
```diff
@@ -0,0 +1,325 @@
...
+            EnableOrganizationDomainsProtection, PhishThresholdLevel,
+            HonorDmarcPolicy, DmarcQuarantineAction, DmarcRejectAction,
+            EnableFirstContactSafetyTips, EnableSimilarUsersSafetyTips,
+            EnableSimilarDomainsSafetyTips, EnableUnusualCharactersSafetyTips,
+            TargetedUserQuarantineTag, TargetedDomainQuarantineTag,
+            MailboxIntelligenceQuarantineTag, SpoofQuarantineTag)
+    }
+    catch {
+        Write-PSFMessage "Failed to retrieve anti-phishing policies: $_" -Tag Test -Level Warning
+        Add-ZtTestResultDetail -SkippedBecause NotConnectedExchange
```
alexandair: Do not report post-check cmdlet failures as “not connected”

The `Service = ('ExchangeOnline')` metadata already prevents the test from starting when Exchange Online is not connected.

If either Exchange cmdlet fails after that pre-check, the cause could instead be missing permissions, an expired session, a transient service issue, or an Exchange cmdlet error. Returning `NotConnectedExchange` hides that distinction.

**Suggested fix:** log the error and return an `Investigate` result that says the anti-phishing policies or rules could not be retrieved. Do not use `NotConnectedExchange` inside these catches.

src/powershell/tests/Test-Assessment.41033.ps1
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    Write-ZtProgress -Activity $activity -Status 'Querying anti-phishing rules'
+    try {
+        $rules = @(Get-AntiPhishRule -ErrorAction Stop | Select-Object `
+            Name, AntiPhishPolicy, Priority, State, RecipientDomainIs, SentTo, SentToMemberOf)
+    }
+    catch {
+        Write-PSFMessage "Failed to retrieve anti-phishing rules: $_" -Tag Test -Level Warning
+        Add-ZtTestResultDetail -SkippedBecause NotConnectedExchange
```
alexandair: Do not report post-check cmdlet failures as “not connected”

The `Service = ('ExchangeOnline')` metadata already prevents the test from starting when Exchange Online is not connected.

If either Exchange cmdlet fails after that pre-check, the cause could instead be missing permissions, an expired session, a transient service issue, or an Exchange cmdlet error. Returning `NotConnectedExchange` hides that distinction.

**Suggested fix:** log the error and return an `Investigate` result that says the anti-phishing policies or rules could not be retrieved. Do not use `NotConnectedExchange` inside these catches.

src/powershell/tests/Test-Assessment.41033.ps1
```diff
@@ -0,0 +1,325 @@
...
+                [void]$reasons.Add('Fail (spoof off)')
+            }
+            if ($policy.HonorDmarcPolicy -ne $true) {
+                [void]$reasons.Add('Fail (dmarc not honored)')
+            }
+            if ($policy.AuthenticationFailAction -ne 'Quarantine') {
+                [void]$reasons.Add('Fail (auth-fail delivered)')
+            }
+
+            # Per-verdict action checks — only when the feature is enabled; name the offending property
```
alexandair, line 212: **Require `Quarantine` for enabled impersonation protections**

The code only fails when an action is `NoAction` or `MoveToJmf`. That means another non-quarantine action, such as redirecting the message, would be treated as compliant.

The specification requires these actions to be `Quarantine`:

- `TargetedUserProtectionAction`
- `TargetedDomainProtectionAction`
- `MailboxIntelligenceProtectionAction`

**Suggested fix:** when the matching protection is enabled, fail whenever its action is not `Quarantine`.

src/powershell/tests/Test-Assessment.41033.ps1
```diff
@@ -0,0 +1,325 @@
...
+    $policyResults = foreach ($policy in $policies) {
+        $isBuiltIn = $policy.IsBuiltInProtection -eq $true
+        $isDefault = $policy.IsDefault -eq $true
+        $reasons   = [System.Collections.Generic.List[string]]::new()
+
+        if (-not $isBuiltIn) {
+            $userCount   = if ($policy.TargetedUsersToProtect)   { @($policy.TargetedUsersToProtect).Count }   else { 0 }
+            $domainCount = if ($policy.TargetedDomainsToProtect) { @($policy.TargetedDomainsToProtect).Count } else { 0 }
+
+            # Impersonation checks — custom policies only; default policy is evaluated against the minimum subset only
```
alexandair, line 156: **Apply the default-policy minimum subset exactly as specified**

The specification says the default policy only needs the spoof, DMARC, and mailbox-intelligence subset. The code additionally requires `AuthenticationFailAction = Quarantine` for the default policy.

That is stricter than the stated rule and could cause an unexpected failure.

**Suggested fix:** evaluate `AuthenticationFailAction` only for custom, in-scope policies, unless the specification is intentionally updated to require it for the default policy too.

src/powershell/tests/Test-Assessment.41033.ps1
```diff
@@ -0,0 +1,325 @@
...
+
+    #region Data Collection
+    Write-PSFMessage '🟦 Start' -Tag Test -Level VeryVerbose
+
+    $activity = 'Checking anti-phishing policy configuration'
+
+    Write-ZtProgress -Activity $activity -Status 'Querying anti-phishing policies'
+    try {
+        $policies = @(Get-AntiPhishPolicy -ErrorAction Stop | Select-Object `
+            Identity, IsBuiltInProtection, IsDefault, Enabled,
```
alexandair, line 39: **Either enforce or remove the unimplemented baseline requirements**

Data is collected, but not evaluated.

The specification lists several required settings, but the code never checks them:

- `EnableUnauthenticatedSender`
- `EnableViaTag`
- `DmarcQuarantineAction`
- `DmarcRejectAction`
- safety-tip settings

This leaves a gap between the documented recommended baseline and the actual compliance test.

**Suggested fix:** decide which of these settings are truly required for this assessment. Then:

1. Add deterministic result reasons to the specification.
2. Check them in the test.
3. Add tests for compliant and non-compliant values.

If they are informational only, remove them from the required baseline in the specification.

src/powershell/tests/Test-Assessment.41033.ps1
```diff
@@ -0,0 +1,325 @@
...
+                @{ Enabled = $policy.EnableTargetedUserProtection;        Prop = 'TargetedUserProtectionAction' }
+                @{ Enabled = $policy.EnableTargetedDomainsProtection;     Prop = 'TargetedDomainProtectionAction' }
+                @{ Enabled = $policy.EnableMailboxIntelligenceProtection; Prop = 'MailboxIntelligenceProtectionAction' }
+            )) {
+                if ($check.Enabled -eq $true -and $policy.($check.Prop) -in @('NoAction', 'MoveToJmf')) {
+                    [void]$reasons.Add("Fail (impersonation action not enforced: $($check.Prop))")
+                }
+            }
+
+            # Quarantine tag self-release — name the offending property
```
alexandair: **Clarify quarantine-tag validation**

The code only flags `DefaultFullAccessPolicy` and `DefaultFullAccessWithNotificationPolicy`. Any other tag is treated as safe.

That is not enough to prove a custom quarantine policy prevents recipient self-release. Conversely, Microsoft supports custom quarantine policies that have no recipient access, so requiring the tag name to be exactly `AdminOnlyAccessPolicy` would be unnecessarily strict.

**Suggested fix:** either:

- query the assigned quarantine-policy configuration and fail when it allows recipient self-release, or
- explicitly limit this test to the built-in tag names and state that limitation in the specification.

src/powershell/tests/Test-Assessment.41033.md
```diff
@@ -0,0 +1,24 @@
...
+**Remediation action**
+
+To configure an anti-phishing policy with impersonation and spoof protection:
+1. Sign in to [Microsoft 365 Defender](https://security.microsoft.com) as a Global Administrator or Security Administrator.
+2. Navigate to **Email & collaboration > Policies & rules > Threat policies > Anti-phishing**.
+3. Select an existing custom policy or click **Create** to add a new one.
+4. Under **Phishing threshold & protection**, enable **Spoof intelligence**, **Mailbox intelligence**, and **Mailbox intelligence protection**.
+5. Enable **Enable users to protect** and add key executives, board members, and finance users to the protected user list.
+6. Enable **Enable domains to protect**, include your own domains, and optionally add partner domains.
+7. Set **Phishing email threshold** to **2 - Aggressive** (Standard) or **3 - More aggressive** (Strict).
```
alexandair: Microsoft’s recommended-settings documentation confirms:

- **Standard:** `3 - More aggressive`
- **Strict:** `4 - Most aggressive`

[Recommended email and collaboration threat policy settings for cloud organizations — Phishing email thresholds](https://learn.microsoft.com/en-us/defender-office-365/recommended-settings-for-eop-and-office365#microsoft-defender-for-office-365-security)

alexandair (changes requested):

src/powershell/tests/Test-Assessment.41033.ps1
```diff
@@ -0,0 +1,393 @@
...
+            if ($policy.EnableViaTag -ne $true) {
+                [void]$reasons.Add('Fail (via tag off)')
+            }
+            if ($policy.EnableFirstContactSafetyTips -ne $true) {
+                [void]$reasons.Add('Fail (first contact safety tips off)')
+            }
+            if ($policy.HonorDmarcPolicy -eq $true -and $policy.DmarcQuarantineAction -ne 'Quarantine') {
+                [void]$reasons.Add('Fail (DMARC quarantine action not quarantine)')
+            }
+            if ($policy.HonorDmarcPolicy -eq $true -and $policy.DmarcRejectAction -ne 'Quarantine') {
```
alexandair: **`DmarcRejectAction` is checked against the wrong value**

The test fails unless `DmarcRejectAction` is `Quarantine`:

```powershell
$policy.DmarcRejectAction -ne 'Quarantine'
```

However, the specification and the remediation guidance both require `DmarcRejectAction` to be `Reject`. As written, a correctly configured policy will fail this test.

**Suggested fix:** compare `DmarcRejectAction` to `Reject` and update the result wording to say that the DMARC reject action is not `Reject`.

src/powershell/tests/Test-Assessment.41033.ps1
```diff
@@ -0,0 +1,393 @@
...
+            if ($policy.EnableSpoofIntelligence -ne $true) {
+                [void]$reasons.Add('Fail (spoof off)')
+            }
+            if ($policy.HonorDmarcPolicy -ne $true) {
+                [void]$reasons.Add('Fail (dmarc not honored)')
+            }
+            if (-not $isDefault -and $policy.AuthenticationFailAction -ne 'Quarantine') {
+                [void]$reasons.Add('Fail (auth-fail delivered)')
+            }
+            if ($policy.EnableUnauthenticatedSender -ne $true) {
```
alexandair: The default policy is checked against controls outside its required subset

The specification says the default policy only needs the spoof, DMARC-honoring, and mailbox-intelligence subset.

However, the new checks apply to the default policy as well:

- `EnableUnauthenticatedSender`
- `EnableViaTag`
- `EnableFirstContactSafetyTips`
- `DmarcQuarantineAction`
- `DmarcRejectAction`

This can fail a tenant whose default policy meets the documented minimum subset, even though a compliant custom policy is applied through an enabled rule.

**Suggested fix:** apply these extra baseline checks only when `-not $isDefault`, unless the specification is updated to make them part of the default-policy subset.

src/powershell/tests/Test-Assessment.41033.ps1
```diff
@@ -0,0 +1,393 @@
...
+
+    #region Data Collection
+    Write-PSFMessage '🟦 Start' -Tag Test -Level VeryVerbose
+
+    $activity = 'Checking anti-phishing policy configuration'
+
+    Write-ZtProgress -Activity $activity -Status 'Querying anti-phishing policies'
+    try {
+        $policies = @(Get-AntiPhishPolicy -ErrorAction Stop | Select-Object `
+            Identity, IsBuiltInProtection, IsDefault, Enabled,
```
alexandair, line 39: **The required `Enabled` setting is not evaluated**

The specification lists `Enabled = True` as a required baseline property. The implementation collects and displays this value, but never uses it to determine compliance.

A policy with `Enabled = False` can therefore receive a `Pass` row result.

**Suggested fix:** add a failure reason when an in-scope custom policy is disabled. Add the reason to the specification’s canonical-result list as well.

src/powershell/tests/Test-Assessment.41033.ps1
```diff
@@ -0,0 +1,393 @@
...
+        return
+    }
+
+    Write-ZtProgress -Activity $activity -Status 'Querying quarantine policies'
+    # $quarantineTagMap: tag name → $true when PermissionToRelease (bit 32) is set (recipients can self-release)
+    # Falls back to $null on error; assessment then uses built-in tag name list as a fallback
+    $quarantineTagMap = $null
+    try {
+        $quarantineTagMap = @{}
+        Get-QuarantinePolicy -ErrorAction Stop | ForEach-Object {
```
alexandair, line 90: **A failed quarantine-policy query can still produce a false pass**

The new `Get-QuarantinePolicy` lookup is a good improvement. However, if it fails, the fallback only recognizes two built-in self-release tag names. A custom quarantine tag that allows recipients to release messages is then treated as compliant. See Test-Assessment.41033.ps1:84-97 and Test-Assessment.41033.ps1:204-215.

**Suggested fix:** if quarantine policies cannot be retrieved and any referenced quarantine tag is custom, return `Investigate` rather than passing the policy. The test cannot confirm whether that custom policy allows recipient self-release.

alexandair (changes requested):

**Also, specification alignment needed**

The new checks add result strings that are not listed in the specification’s required canonical strings, including the safety-tip, Via-tag, unauthenticated-sender, and DMARC-action reasons. See Test-Assessment.41033.ps1:149-192.

Also, the implementation now treats a threshold below 3 as `Investigate`, while the specification still defines only threshold 1 as `Investigate` and lists level 2 as the Standard recommendation.

**Suggested fix:** update 41033.md in the same change to define the new baseline and exact reason strings, including the current Microsoft threshold recommendations.

src/powershell/tests/Test-Assessment.41033.ps1
```diff
@@ -0,0 +1,410 @@
...
+        }
+        Add-ZtTestResultDetail @params
+        return
+    }
+
+    Write-ZtProgress -Activity $activity -Status 'Querying quarantine policies'
+    # $quarantineTagMap: tag name → $true when recipients can self-release without notification
+    # Falls back to $null on error; assessment then uses built-in tag name list as a fallback
+    $quarantineTagMap = $null
+    try {
```
alexandair, line 88: **Quarantine-policy validation cannot detect self-release**

Test-Assessment.41033.ps1 reads `EndUserQuarantinePermissionsValue` from `Get-QuarantinePolicy` and tests bit `32`.

This does not reliably evaluate the policy:

- `EndUserQuarantinePermissionsValue` is a write-side parameter for `New-QuarantinePolicy` and `Set-QuarantinePolicy`; it is not returned by `Get-QuarantinePolicy`.
- When the missing value is cast to `int`, it becomes `0`, so `$canSelfRelease` is always `$false`.
- Even if the integer were present, bit `32` represents `PermissionToAllowSender`, not `PermissionToRelease`. The release bit is `4`.

As a result, a referenced `DefaultFullAccessPolicy` or custom quarantine policy that permits release can be marked compliant whenever `Get-QuarantinePolicy` succeeds. The fallback list does not help, because it is only used when the entire cmdlet call fails.

This is corroborated by the established quarantine-policy implementation guidance in ztspecs/specs/secops/41030.md and its documented bit mapping in ztspecs/specs/secops/41030.md. Microsoft’s quarantine-policy documentation also identifies `EndUserQuarantinePermissionsValue` as the configuration value used by the create/update cmdlets.

**Required change:** build the map from `EndUserQuarantinePermissions`, for example by testing whether its formatted value contains `PermissionToRelease: True`. Then apply the intended notification rule from the 41033 specification. Add cases for:

- `DefaultFullAccessPolicy`;
- `DefaultFullAccessWithNotificationPolicy`;
- a custom policy with release enabled;
- a custom policy without release permission;
- an unavailable quarantine-policy query.

src/powershell/tests/Test-Assessment.41033.ps1
```diff
@@ -0,0 +1,410 @@
...
+            PhishThreshold  = $threshDisplay
+            Result          = $rowResult
+            RowStatus       = $rowStatus
+        }
+    }
+    $policyResults = @($policyResults)
+
+    # Identify in-scope policy groups
+    $defaultResult   = $policyResults | Where-Object { $_.IsDefault } | Select-Object -First 1
+    $ruleRefPolicies = @($policyResults | Where-Object { -not $_.IsBuiltIn -and -not $_.IsDefault -and $_.Scope -ne 'Not applied' })
```
alexandair, line 315: **The default policy is treated as compliant when it is missing.**  

   Test-Assessment.41033.ps1 sets `$defaultMeetsSubset` to true when no default-policy result exists. A partial or malformed cmdlet response can therefore pass if a custom policy passes, even though the required default-policy subset was never checked.

   **Suggested fix:** return `Investigate` when no default policy is present, similarly to the existing zero-row handling.

src/powershell/tests/Test-Assessment.41033.ps1
```diff
@@ -0,0 +1,410 @@
...
+        $reasons   = [System.Collections.Generic.List[string]]::new()
+        $unverifiedCustomQuarantineTags = [System.Collections.Generic.List[string]]::new()
+
+        if (-not $isBuiltIn) {
+            $userCount   = if ($policy.TargetedUsersToProtect)   { @($policy.TargetedUsersToProtect).Count }   else { 0 }
+            $domainCount = if ($policy.TargetedDomainsToProtect) { @($policy.TargetedDomainsToProtect).Count } else { 0 }
+
+            # Impersonation checks — custom policies only; default policy is evaluated against the minimum subset only
+            if (-not $isDefault) {
+                if ($policy.Enabled -ne $true) {
```
alexandair, line 158: **`Fail (policy disabled)` is not a canonical result reason.**  

   The code emits this value in Test-Assessment.41033.ps1:145-148, but it is absent from the canonical reason list in ztspecs/specs/secops/41033.md.

   **Suggested fix:** add the exact reason and its deterministic ordering to the specification, or avoid emitting it. The baseline does require `Enabled = True`, so adding it to the spec is preferable.

src/powershell/tests/Test-Assessment.41033.ps1
```diff
@@ -0,0 +1,410 @@
...
+            else {
+                $aggregation        = 'no custom anti-phishing policy with full impersonation controls is applied to any recipient'
+                $testResultMarkdown = "⚠️ Only the default anti-phishing policy is in effect. The default policy is evaluated against the minimum spoof and DMARC subset only; it does not enforce targeted-user, domain, or org-domain impersonation protection. Configure a custom policy with impersonation controls and assign it via an enabled rule.`n`n%TestResult%"
+            }
+        }
+    }
+    else {
+        $passed             = $true
+        $aggregation        = 'at least one compliant policy applies, default policy meets required subset, no orphaned rules'
+        $testResultMarkdown = "✅ Anti-phishing policies are configured with spoof intelligence, DMARC honoring, mailbox intelligence, and impersonation protection for users and domains; impersonation and spoof verdicts are quarantined with end-user notification.`n`n%TestResult%"
```
alexandair: **The pass message can claim notification where none exists.**

The specification accepts AdminOnlyAccessPolicy as a stricter configuration, but Test-Assessment.41033.ps1:352 always says verdicts are quarantined “with end-user notification.” The accompanying remediation also directs customers to AdminOnlyAccessPolicy in Test-Assessment.41033.md:13.

Suggested fix: use outcome-neutral wording such as “quarantined without unapproved self-release,” or report the assigned quarantine-policy behavior accurately.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.41033.ps1
Manoj-Kesana: @Trackmunkey - can you please add this to the spec

src/powershell/tests/Test-Assessment.41033.ps1
```diff
@@ -88,7 +88,7 @@ function Test-Assessment-41033 {
     try {
         $quarantineTagMap = @{}
         Get-QuarantinePolicy -ErrorAction Stop | ForEach-Object {
-            $canSelfRelease = (([int]$_.EndUserQuarantinePermissionsValue -band 32) -ne 0)
+            $canSelfRelease = $_.EndUserQuarantinePermissions -match 'PermissionToRelease:\s+True'
```
alexandair: **Avoid parsing the rendered permissions object.** 

Test uses `-match` against `EndUserQuarantinePermissions`. This relies on that Exchange object's `ToString()`/remoting representation retaining the exact text `PermissionToRelease: True`; formatting differences (notably whitespace around `:`) cause a false negative and allow a self-release policy to pass. Read the Boolean property directly, e.g. `EndUserQuarantinePermissions.PermissionToRelease -eq $true`, rather than regex-matching display text.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.41033.ps1
Manoj-Kesana: resolved

alexandair (approved):
LGTM

PR #1398, opened 2026-07-13
SecOps - 41114 - Microsoft Teams protection policies are configured to block malicious messages, links, and attachments

praneeth-0000:


src/powershell/tests/Test-Assessment.41114.ps1
```diff
@@ -0,0 +1,187 @@
...
+    param()
+
+    #region Data Collection
+    Write-PSFMessage '🟦 Start' -Tag Test -Level VeryVerbose
+    $mgEnv = (Get-MgContext).Environment
+    if ($mgEnv -eq 'USGov' -or $mgEnv -eq 'USGovDoD') {
+        Write-PSFMessage 'This test is not applicable to the USGov or USGovDoD environments.' -Tag Test -Level VeryVerbose
+        Add-ZtTestResultDetail -SkippedBecause NotSupported -Result 'This test is not applicable to the USGov or USGovDoD environments.'
+        return
+    }
```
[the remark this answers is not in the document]
praneeth-0000: Spec explicitly mentions to skip US Gov tenants as Teams user reporting is unavailable.

alexandair:
**ZAP exceptions are not collected or assessed.** 
 
   The spec requires `Investigate` when Teams ZAP exceptions are present, but the implementation only queries `Get-TeamsProtectionPolicy` and evaluates `ZapEnabled`. It never calls `Get-TeamsProtectionPolicyRule` or evaluates `ExceptIfSentTo`, `ExceptIfSentToMemberOf`, and `ExceptIfRecipientDomainIs`. Therefore, a tenant with ZAP enabled but a broad exclusion rule is incorrectly reported as **Pass**. Microsoft’s documented Teams configuration uses the rule cmdlet for these exceptions.

alexandair:
**Safe Links can pass based on an unapplied or overridden policy.**  

   The implementation reports pass if *any* `Get-SafeLinksPolicy` object enables Teams. A Safe Links policy created through PowerShell is not applied until associated with a Safe Links rule, and custom policy rules determine recipient scope and precedence. Consequently, an unassigned enabled policy—or the built-in policy alongside an applicable custom policy with Teams disabled—can make this test pass incorrectly. The implementation must query `Get-SafeLinksRule`, restrict evaluation to enabled, applicable/in-scope policies, and account for custom-policy precedence.

alexandair:
**Ambiguous Boolean values are classified as failures.**  

   The specification requires `Investigate` when Q1–Q4 produce ambiguous values. For Q1, Q2, and Q3, `$null` property values become `$false` via the `-eq $true` checks, resulting in a definite **Fail**. Only Q4 explicitly handles an unknown value. For example, a returned Teams protection policy with `ZapEnabled = $null` is reported as “❌ No,” rather than “⚠️ Investigate.” Treat null/ambiguous properties as unknown before setting the fail flag.

praneeth-0000:
> **Safe Links can pass based on an unapplied or overridden policy.**
> 
> The implementation reports pass if _any_ `Get-SafeLinksPolicy` object enables Teams. A Safe Links policy created through PowerShell is not applied until associated with a Safe Links rule, and custom policy rules determine recipient scope and precedence. Consequently, an unassigned enabled policy—or the built-in policy alongside an applicable custom policy with Teams disabled—can make this test pass incorrectly. The implementation must query `Get-SafeLinksRule`, restrict evaluation to enabled, applicable/in-scope policies, and account for custom-policy precedence.

Hi @astaykov , I think the logic similar to test 41032 needs to be implemented for this test as well? Currently, in this test there is no mention of using `Get-SafeLinksRule`. What do you think?

src/powershell/tests/Test-Assessment.41114.ps1
```diff
@@ -0,0 +1,334 @@
...
+    $safeLinksRules = $null
+    $q3rError = $null
+    try {
+        # Q3b: Retrieve Safe Links rules; only enabled rules determine which policies are in scope.
+        $safeLinksRules = @(Get-SafeLinksRule -ErrorAction Stop | Select-Object Name, SafeLinksPolicy, State)
+    }
+    catch {
+        $q3rError = $_
+        Write-PSFMessage "Q3r: Failed to retrieve Safe Links rules: $_" -Tag Test -Level Warning
+    }
```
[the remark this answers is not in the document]
praneeth-0000: fixed

src/powershell/tests/Test-Assessment.41114.ps1
```diff
@@ -0,0 +1,338 @@
...
+            $anyFail = $true
+            $controlRows.Add([PSCustomObject]@{
+                Setting        = "[Safe Links for Teams]($q3PortalLink)"
+                Policy         = ''
+                AppliedViaRule = ''
+                Enabled        = '❌ No enabled rule'
+                Status         = '❌ Fail — no enabled Safe Links rule'
+            })
+        }
+        else {
```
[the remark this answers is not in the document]
praneeth-0000: resolved

src/powershell/tests/Test-Assessment.41114.ps1
```diff
@@ -0,0 +1,362 @@
...
+    # --- Q3: Safe Links for Teams — evaluated per enabled Safe Links rule/policy ---
+    # Build lookup: policy Identity → policy object.
+    $safeLinksPolicyByIdentity = @{}
+    if ($safeLinksPolicies) {
+        foreach ($p in $safeLinksPolicies) {
+            $safeLinksPolicyByIdentity[$p.Identity] = $p
+        }
+    }
+
+    # Zero or null $safeLinksPolicies is a retrieval/permission anomaly: every MDO P1 tenant has at
```
alexandair, line 214: **No enabled Safe Links rule can still produce Pass or Investigate instead of Fail.**

   The specification explicitly requires **Fail** when no Safe Links rule is enabled and gives Fail precedence over Investigate ([spec requirement](https://github.com/microsoft/ztspecs/blob/main/specs/secops/41114.md#L101-L105)).

   However, the implementation:
   - skips rule evaluation entirely when the policy query is empty, producing Investigate ([policy-anomaly branch](#L229-L241));
   - evaluates Built-in Protection when there are no enabled rules and can report Pass ([built-in fallback](#L243-L270)).

   Thus, a tenant with zero enabled Safe Links rules can avoid the required failure. Rule availability/count should be evaluated independently, and a successful rule query returning zero enabled rules must set `$anyFail`.

src/powershell/tests/Test-Assessment.41114.ps1
```diff
@@ -0,0 +1,362 @@
...
+        $controlRows.Add([PSCustomObject]@{
+            Setting        = "[ZAP for Teams]($q2PortalLink)"
+            Policy         = ''
+            AppliedViaRule = ''
+            Enabled        = '⚠️ Unknown'
+            Status         = '⚠️ Investigate — could not retrieve Teams Protection rules'
+        })
+    }
+    else {
+        $enabledTeamsRules = @($teamsProtectionPolicyRule | Where-Object { $_.State -eq 'Enabled' })
```
alexandair: **Ambiguous rule states are treated as “not enabled.”**

   Both rule sets retain only `State == 'Enabled'` and then treat an empty result as no enabled rule ([Teams rules](#L156-L158)). A returned rule with a null or unrecognized `State` is therefore classified as an explicit absence of enabled rules rather than an ambiguous Q2/Q3 value.

   The specification requires ambiguous Q1–Q4 values to produce Investigate unless another definite failure exists ([ambiguity requirement](https://github.com/microsoft/ztspecs/blob/main/specs/secops/41114.md#L102-L105)). Explicitly distinguish `Disabled` from null/unrecognized rule states.

src/powershell/tests/Test-Assessment.41114.ps1
```diff
@@ -0,0 +1,362 @@
...
+            $controlRows.Add([PSCustomObject]@{
+                Setting        = "[ZAP for Teams]($q2PortalLink)"
+                Policy         = ''
+                AppliedViaRule = ''
+                Enabled        = '❌ No enabled rule'
+                Status         = '❌ Fail — no enabled Teams Protection rule'
+            })
+        }
+        else {
+            foreach ($rule in $enabledTeamsRules) {
```
alexandair: The output specification requires one row per distinct referenced policy, with all associated rule names. The implementation emits one row per rule, creating duplicate policy rows when multiple rules reference the same policy.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.41114.ps1
praneeth-0000: Hi Aleks, There can only be 1 Teams Protection policy rule per tenant. Regarding SafeLinks rule, I've already tested attaching a rule to an existing policy already attached with rule and I got this error `New-SafeLinksRule: ||Policy "xyzPolicy" already has rule "xyzRule" associated with it.` and I wasn't allowed to attach another rule.

src/powershell/tests/Test-Assessment.41114.ps1
praneeth-0000: Both `Get-TeamsProtectionPolicyRule` and `Get-SafeLinksRule` has `RuleState` type for `State` property and it can only have Enabled or Disabled members. So, practically we'd only be getting those 2 values.

src/powershell/tests/Test-Assessment.41114.ps1
praneeth-0000: This is fixed, acc. to new spec update.

src/powershell/tests/Test-Assessment.41114.ps1
```diff
@@ -0,0 +1,347 @@
...
+            }
+        }
+        else {
+            $teamsProtectionRule = $teamsProtectionPolicyRule | Select-Object -First 1
+            if ($teamsProtectionRule.State -eq 'Enabled') {
+                $enabledTeamsRule = $teamsProtectionRule
+                $hasExceptions = $teamsProtectionRule.ExceptIfSentTo -or
+                    $teamsProtectionRule.ExceptIfSentToMemberOf -or
+                    $teamsProtectionRule.ExceptIfRecipientDomainIs
+            }
```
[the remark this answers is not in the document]
praneeth-0000: not necessary, there is only 1 rule.

src/powershell/tests/Test-Assessment.41114.ps1
```diff
@@ -0,0 +1,348 @@
...
+            $controlRows.Add([PSCustomObject]@{
+                Setting        = "[Safe Links for Teams]($q3PortalLink)"
+                Policy         = ''
+                AppliedViaRule = ''
+                Enabled        = '⚠️ Unknown'
+                Status         = '⚠️ Investigate — could not retrieve Safe Links rules'
+            })
+        }
+        else {
+            $enabledSafeLinksRules = @($safeLinksRules | Where-Object { $_.State -eq 'Enabled' })
```
alexandair, line 280: **Unknown Safe Links rule states can yield a false Pass.**  

   Test retains only rules whose `State` is exactly `Enabled`; null or unrecognized states are silently treated as disabled. Therefore, if Built-in Protection is enabled, Q3 can pass even though rule applicability is ambiguous. The specification requires ambiguous Q1–Q4 data to yield `Investigate` unless a failure exists. Explicitly identify `Disabled` as the only non-ambiguous inactive state and set `$anyInvestigate` for other values. See 41114:101-105.md.

src/powershell/tests/Test-Assessment.41114.ps1
```diff
@@ -0,0 +1,348 @@
...
+            foreach ($rule in $enabledSafeLinksRules) {
```
alexandair, line 281: **Duplicate policy rows:**

 Multiple enabled rules pointing to the same policy produce multiple rows rather than one distinct-policy row containing all rule names, contrary to 41114.md:112-118.

src/powershell/tests/Test-Assessment.41114.ps1
```diff
@@ -0,0 +1,348 @@
...
+            if (-not $anyFail) {
+                $anyInvestigate = $true
+                $zapStatus = '⚠️ Investigate — could not retrieve Teams Protection rules to check exclusions'
+            }
+        }
+        else {
+            $teamsProtectionRule = $teamsProtectionPolicyRule | Select-Object -First 1
+            if ($teamsProtectionRule.State -eq 'Enabled') {
+                $enabledTeamsRule = $teamsProtectionRule
+                $hasExceptions = $teamsProtectionRule.ExceptIfSentTo -or
```
alexandair: **ZAP exception status omits the exception property:** 

Test reports a generic exception rather than naming `ExceptIfSentTo`, `ExceptIfSentToMemberOf`, or `ExceptIfRecipientDomainIs`, as required by 41114.md:112-118.

alexandair (changes requested):
@praneeth-0000 Please, address my review.

src/powershell/tests/Test-Assessment.41114.ps1
praneeth-0000: fixed

src/powershell/tests/Test-Assessment.41114.ps1
praneeth-0000: There is a 1-1 relation between policy and rule. So 2 rules can't reference 1 policy.

src/powershell/tests/Test-Assessment.41114.ps1
praneeth-0000: fixed

alexandair (approved):
LGTM

PR #1399, opened 2026-07-13
add security reader note in readme for test 41114

praneeth-0000:
This PR add a Note mentioning some Exchange cmdlets need Security Reader role to work properly. Ex:- 41114 Spec's `Get-ReportSubmissionPolicy`


alexandair (approved):
LGTM

PR #1407, opened 2026-07-15
SecOps - 41034 - Anti-spam (hosted content filter) policies are configured with recommended thresholds and actions

sandeepjha000:
[Spec-41034](https://github.com/microsoft/ztspecs/blob/main/specs/secops/41034.md)
[Issue-906](https://github.com/microsoft/ztspecs/issues/906)

### DIFF
--- a/src/powershell/tests/Test-Assessment.41034.ps1
+++ b/src/powershell/tests/Test-Assessment.41034.ps1
@@ -121,7 +121,6 @@ function Test-Assessment-41034 {
 
     $passed         = $true
     $hasInvestigate = $false
-    $hasAllowedList = $false
 
     $policyRows = [System.Collections.Generic.List[PSCustomObject]]::new()
 
@@ -134,25 +133,11 @@ function Test-Assessment-41034 {
             # Orphan rule: enabled rule references a policy not found in Get-HostedContentFilterPolicy → Investigate
             $hasInvestigate = $true
             $policyRows.Add([PSCustomObject]@{
-                Identity                         = $identity
-                IsDefault                        = $false
-                IsOrphan                         = $true
-                RuleName                         = $ruleName
-                FailReasons                      = @()
-                RowResult                        = 'Investigate'
-                SpamAction                       = $null
-                HighConfidenceSpamAction         = $null
-                PhishSpamAction                  = $null
-                HighConfidencePhishAction        = $null
-                BulkSpamAction                   = $null
-                BulkThreshold                    = $null
-                MarkAsSpamBulkMail               = $null
-                PhishZapEnabled                  = $null
-                SpamZapEnabled                   = $null
-                InlineSafetyTipsEnabled          = $null
-                AllowedSendersCount              = $null
-                AllowedSenderDomainsCount        = $null
-                HighConfidencePhishQuarantineTag = $null
+                Identity  = $identity
+                IsDefault = $false
+                IsOrphan  = $true
+                RuleName  = $ruleName
+                RowResult = 'Investigate'
             })
             continue
         }
@@ -180,9 +165,6 @@ function Test-Assessment-41034 {
         $rowFails = $failReasons.Count -gt 0
         if ($rowFails) {
             $passed = $false
-            if ($allowedSendersCount -gt 0 -or $allowedSenderDomainsCount -gt 0) {
-                $hasAllowedList = $true
-            }
         }
 
         $rowResult       = if ($rowFails) { 'Fail' } else { 'Pass' }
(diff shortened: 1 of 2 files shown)


src/powershell/tests/Test-Assessment.41034.ps1
```diff
@@ -0,0 +1,298 @@
...
+
+            # Bulk & ZAP: Bulk: <n>/6 • MarkBulk: Y/N • ZAP: Y/N • SafetyTips: Y/N
+            # ZAP is a single Y/N: Y only when both PhishZapEnabled AND SpamZapEnabled are true
+            $zapDisplay        = if ($row.PhishZapEnabled -eq $true -and $row.SpamZapEnabled -eq $true) { 'Y' } else { 'N' }
+            $markBulkDisplay   = if ($row.MarkAsSpamBulkMail -eq 'On') { 'Y' } else { 'N' }
+            $safetyTipsDisplay = if ($row.InlineSafetyTipsEnabled -eq $true) { 'Y' } else { 'N' }
+            $bulkZapDisplay    = "Bulk: $($row.BulkThreshold)/6 • MarkBulk: $markBulkDisplay • ZAP: $zapDisplay • SafetyTips: $safetyTipsDisplay"
+
+            # Allow lists & quarantine tag: Senders: N • Domains: N • HC-Phish tag: <tag>
+            $allowListsDisplay = "Senders: $($row.AllowedSendersCount) • Domains: $($row.AllowedSenderDomainsCount) • HC-Phish tag: $($row.HighConfidencePhishQuarantineTag)"
```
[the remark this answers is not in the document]
sandeepjha000: resolved in commit: [453ad12](https://github.com/microsoft/zerotrustassessment/pull/1407/commits/453ad12105e31f88768e1d58e764d0978abd0a86)

src/powershell/tests/Test-Assessment.41034.ps1
```diff
@@ -0,0 +1,298 @@
...
+            AllowedSenderDomainsCount        = $allowedSenderDomainsCount
+            HighConfidencePhishQuarantineTag = $policy.HighConfidencePhishQuarantineTag
+        })
+    }
+
+    # Aggregate verdict: Fail > Investigate > Pass
+    $customStatus = $null
+    if (-not $passed) {
+        $testResultMarkdown = "❌ One or more anti-spam policies allow high-confidence phishing to be delivered (action is not Quarantine), set a permissive bulk threshold, or contain entries in the per-policy allowed-sender list that bypass spam, spoofing, and most phishing filtering for those senders. Operational exceptions for legitimate vendors belong in the **[Tenant Allow/Block List](https://security.microsoft.com/tenantAllowBlockList)**, where they can be reviewed and expired.`n`n%TestResult%"
+    }
```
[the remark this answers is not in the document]
sandeepjha000: NA - Fail message is as per spec

src/powershell/tests/Test-Assessment.41034.ps1
```diff
@@ -0,0 +1,298 @@
...
+    # Collect in-scope policy identities: Default first, then all referenced by enabled rules (deduplicated)
+    $inScopeIdentities = [System.Collections.Generic.List[string]]::new()
+    if ($defaultPolicy) {
+        $inScopeIdentities.Add($defaultPolicy.Identity)
+    }
+    foreach ($policyName in $rulesForPolicy.Keys) {
+        if (-not $inScopeIdentities.Contains($policyName)) {
+            $inScopeIdentities.Add($policyName)
+        }
+    }
```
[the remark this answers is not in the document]
sandeepjha000: resolved in commit: [453ad12](https://github.com/microsoft/zerotrustassessment/pull/1407/commits/453ad12105e31f88768e1d58e764d0978abd0a86)

src/powershell/tests/Test-Assessment.41034.ps1
```diff
@@ -0,0 +1,297 @@
...
+    foreach ($rule in $enabledRules) {
+        if (-not $rulesForPolicy.ContainsKey($rule.HostedContentFilterPolicy)) {
+            $rulesForPolicy[$rule.HostedContentFilterPolicy] = $rule.Name
+        }
+    }
+
+    # Default policy (IsDefault == True) — always in-scope as the catch-all
+    $defaultPolicy = $allPolicies | Where-Object { $_.IsDefault -eq $true } | Select-Object -First 1
+
+    # Collect in-scope policy identities: Default first, then all referenced by enabled rules (deduplicated).
```
alexandair, line 126: **A missing Default policy produces a false Pass**

The code finds the Default policy but only adds it when it exists. If policy retrieval returns custom policies but none has `IsDefault = $true`, and there are no enabled rules, `inScopeIdentities` and `policyRows` remain empty. The assessment then returns Pass with an empty table.

This violates the requirement that the Default policy must be evaluated before the test can pass. A missing Default policy is anomalous retrieval data and should return `Investigate`, similar to the zero-policy path.

src/powershell/tests/Test-Assessment.41034.ps1
```diff
@@ -0,0 +1,297 @@
...
+        $allowedSenderDomainsCount = ($policy.AllowedSenderDomains | Measure-Object).Count
+
+        $failReasons = [System.Collections.Generic.List[string]]::new()
+        # Spec-named reasons first (most impactful)
+        if ($policy.HighConfidencePhishAction -ne 'Quarantine')          { [void]$failReasons.Add('HC-phish delivered') }
+        if ($policy.PhishSpamAction -ne 'Quarantine')                    { [void]$failReasons.Add('phish delivered') }
+        if ($policy.HighConfidenceSpamAction -ne 'Quarantine')           { [void]$failReasons.Add('HC-spam delivered') }
+        if ($policy.SpamAction -notin @('MoveToJmf', 'Quarantine'))      { [void]$failReasons.Add("spam action $($policy.SpamAction)") }
+        if ($policy.BulkSpamAction -notin @('MoveToJmf', 'Quarantine')) { [void]$failReasons.Add("bulk action $($policy.BulkSpamAction)") }
+        if ($policy.BulkThreshold -gt 6)                                 { [void]$failReasons.Add("bulk threshold $($policy.BulkThreshold)") }
```
alexandair: **A null `BulkThreshold` incorrectly satisfies the baseline**

The check only fails when `BulkThreshold -gt 6`:

In PowerShell, a null value does not satisfy that expression, so an unavailable or malformed threshold is treated as compliant. The specification requires the value to satisfy `<= 6`; null does not meet that baseline. Explicitly handle null as either Fail or Investigate.

src/powershell/tests/Test-Assessment.41034.ps1
```diff
@@ -0,0 +1,297 @@
...
+            continue
+        }
+
+        # Evaluate each baseline property and collect named fail reasons
+        $allowedSendersCount       = ($policy.AllowedSenders | Measure-Object).Count
+        $allowedSenderDomainsCount = ($policy.AllowedSenderDomains | Measure-Object).Count
+
+        $failReasons = [System.Collections.Generic.List[string]]::new()
+        # Spec-named reasons first (most impactful)
+        if ($policy.HighConfidencePhishAction -ne 'Quarantine')          { [void]$failReasons.Add('HC-phish delivered') }
```
alexandair: Remove all these `[void]`. It's redundant.

src/powershell/tests/Test-Assessment.41034.ps1
```diff
@@ -0,0 +1,297 @@
...
+    $policyByIdentity = @{}
+    foreach ($policy in $allPolicies) {
+        $policyByIdentity[$policy.Identity] = $policy
+    }
+
+    # Spec: zero rows from Get-HostedContentFilterRule is NOT an error — evaluate Default policy alone.
+    $enabledRules = @($allRules | Where-Object { $_.State -eq 'Enabled' })
+
+    # Map: policy identity → rule name (join key: rule.HostedContentFilterPolicy == policy.Identity)
+    $rulesForPolicy = @{}
```
alexandair, line 101: I am not sure if this should be addressed or not.

The code maps each policy identity to a **single** rule name.

For example, given:

| Rule | Policy |
|---|---|
| Executives | Strict Policy |
| Finance | Strict Policy |

The first rule stores:

`Strict Policy → Executives`

When the second rule is processed, `ContainsKey()` is already true, so `Finance` is discarded.

This does **not affect the verdict**: the policy is correctly evaluated once, which matches the specification’s one-row-per-policy output. It only makes the Scope column incomplete because it reports:

`Applied via rule Executives`

instead of something like:

`Applied via rules Executives, Finance`

Therefore, this is a reporting completeness suggestion, not a correctness issue. It could also reasonably be removed from the review because the specification’s singular display format says `Applied via rule <Name>` and does not explicitly require listing every associated rule.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.41034.ps1
sandeepjha000: The default policy always exists based on the spec, but have added extra condition to handle to be extra safe
fixed in commit: [adf6fb3](https://github.com/microsoft/zerotrustassessment/pull/1407/commits/adf6fb30d864bca8a647e6e07ac509b0c35fa4d4)

src/powershell/tests/Test-Assessment.41034.ps1
sandeepjha000: fixed in commit: fixed in commit: [adf6fb3](https://github.com/microsoft/zerotrustassessment/pull/1407/commits/adf6fb30d864bca8a647e6e07ac509b0c35fa4d4)

src/powershell/tests/Test-Assessment.41034.ps1
sandeepjha000: fixed in commit: [adf6fb3](https://github.com/microsoft/zerotrustassessment/pull/1407/commits/adf6fb30d864bca8a647e6e07ac509b0c35fa4d4)

src/powershell/tests/Test-Assessment.41034.ps1
sandeepjha000: makes sense to display all rules. Fixed in commit: [adf6fb3](https://github.com/microsoft/zerotrustassessment/pull/1407/commits/adf6fb30d864bca8a647e6e07ac509b0c35fa4d4)

alexandair (approved):
LGTM

PR #1426, opened 2026-07-20
Implement error sanitization and reporting improvements in tests

### LINKED ISSUE #1422: Failed Graph requests can embed a live bearer token in TestResult
**Title:** Failed Graph requests can embed a live bearer token in TestResult

**Body:**

When a test's `Invoke-MgGraphRequest` call fails with an HTTP error (e.g. a
transient 401 on an endpoint that only supports delegated auth for
app-only callers), the full failed-request object — including the
`Authorization: Bearer <token>` header — gets serialized into that test's
`TestResult` field in the export JSON (`ZeroTrustAssessmentReport.json`).

Observed on module 2.4.0, test 24546 (Windows Automatic Enrollment,
`GET /policies/mobileDeviceManagementPolicies`), which failed with
`401 Unauthorized` / `"Unsupported app-only call"`. The resulting
`TestResult` string contains the complete PowerShell error record,
including `TargetObject.Headers.Authorization`, which is a live bearer
token for the calling app-only identity — decodable (JWT), revealing the
app's display name, appid, oid, and full Graph role list.

**Impact:** `ZeroTrustAssessmentReport.json` is designed to be handled as
sensitive report data (the module's own docs say as much), but this
specific field can additionally contain live credential material — a
different and stricter sensitivity class than "attacker-usable
environment roadmap." Anyone storing/transmitting/displaying this export
downstream needs to redact `TestResult` on `Error`-status tests, or the
export can leak a working token for however long that token's lifetime is
(observed: ~24h in this case).

**Suggested fix:** when catching the request exception for `TestResult`
population, strip the `Authorization` header (and any other credential
material) from the captured request/response object before
stringifying — same as most HTTP client libraries' built-in
credential-scrubbing on exception logging.

**Reproduction:** difficult to force deterministically (requires a
transient/permission-boundary 401 on a live tenant), but any test hitting
an endpoint outside the app's granted scope, or a genuinely transient
Graph error, should reproduce it.

alexandair:
Fix #1422 

This pull request introduces a comprehensive overhaul of error sanitization for PowerShell test diagnostics to prevent sensitive credential data from being persisted in logs, reports, or error records. The main improvements include a new defense-in-depth sanitization function, safe error record/message generation, and updated error formatting and logging logic to ensure only allow-listed diagnostic fields are exposed. Extensive tests are added to validate these protections.

**Error sanitization and redaction:**

* Added `Protect-ZtReportText` function to redact credential-bearing HTTP headers and sensitive values from report content before persistence, and integrated it into `Add-ZtTestResultDetail`.
* Updated test result persistence and logging to always sanitize error details, including at the TestResult persistence boundary and in optional test logs.

**Safe error record and message generation:**

* Introduced `New-ZtSafeErrorRecord` to create sanitized error records with only allow-listed diagnostic fields and without unsafe TargetObject data. This prevents request/response objects containing bearer tokens from being retained in test statistics or passed into logging.
* Added `Get-ZtSafeErrorMessage` to extract safe, structured diagnostic summaries (request method/path, HTTP status, Graph error code, and correlation IDs) from error records, omitting unstructured exception messages, headers, and response bodies.

**Error formatting and reporting:**

* Implemented `Format-ZtTestErrorDetail` to generate markdown-formatted error summaries using only sanitized fields for failed tests. It adds bounded metadata such as exception type, error ID, HTTP status, and source location.
* Refactored error handling in `Invoke-ZtTest.ps1` and `Write-ZtTestError.ps1` to use the new safe error record/message pipeline, ensuring all error reporting is sanitized and consistent.

**Testing:**

* Added `ErrorSanitization.Tests.ps1` with comprehensive tests to verify that credentials and sensitive values are never persisted in logs or reports, and that only safe diagnostic information is exposed.

### DIFF
--- a/code-tests/commands/ErrorSanitization.Tests.ps1
+++ b/code-tests/commands/ErrorSanitization.Tests.ps1
@@ -1,0 +1,138 @@
+Describe "Error sanitization" {
+	BeforeAll {
+		$here = $PSScriptRoot
+		$srcRoot = Join-Path $here "../../src/powershell"
+
+		. (Join-Path $srcRoot "private/core/Protect-ZtReportText.ps1")
+		. (Join-Path $srcRoot "private/core/Get-ZtHttpStatusCode.ps1")
+		. (Join-Path $srcRoot "private/tests/Get-ZtSafeErrorMessage.ps1")
+		. (Join-Path $srcRoot "private/tests/New-ZtSafeErrorRecord.ps1")
+		. (Join-Path $srcRoot "private/tests/Format-ZtTestErrorDetail.ps1")
+		. (Join-Path $srcRoot "private/tests/Write-ZtTestError.ps1")
+		. (Join-Path $srcRoot "private/tests/Write-ZtTestLog.ps1")
+		. (Join-Path $srcRoot "private/core/Add-ZtTestResultDetail.ps1")
+
+		function New-CanaryErrorRecord {
+			$script:canaryToken = 'eyJhbGciOiJub25lIn0.eyJhcHBpZCI6InJlZGFjdGlvbi10ZXN0In0.signature'
+			$request = [System.Net.Http.HttpRequestMessage]::new([System.Net.Http.HttpMethod]::Get, "https://graph.microsoft.com/beta/example?access_token=$script:canaryToken")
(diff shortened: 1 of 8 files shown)


astaykov (approved):
LGTM

Issue #1434, opened 2026-07-21
Audit Logging test is wrong


DeanGross144:
We have the purview audit log enabled, but we failed test 35037 in the AI Pillar

astaykov:
hey @DeanGross144 ,

Thanks for catching this. Looking at internal implementation notes we have identified a blocking issue that prevent us from reliably identifying Purview audit log settings. 
For time being we will be retiring this check from further releases until we figure out more reliable way to perform this check in the context of Zero Trust Assessment package.

DeanGross144:
Thanks for the quick response, I am seeing the same check in the Data Pillar. Is it supposed to be in both pillars?

astaykov:
Yes, It is absolutely the same check. And it appears in both pillars because it is relevant for them both.

astaykov:
We retired this check with version 2.5.4: https://www.powershellgallery.com/packages/ZeroTrustAssessment/2.5.4-preview

AndrewSWX:
> hey [@DeanGross144](https://github.com/DeanGross144) ,
> 
> Thanks for catching this. Looking at internal implementation notes we have identified a blocking issue that prevent us from reliably identifying Purview audit log settings. For time being we will be retiring this check from further releases until we figure out more reliable way to perform this check in the context of Zero Trust Assessment package.

However, looking at the output logs, the correct audit log setting was identified (as true). It just wasn't reflected in the report.

Log extract...

 Test: 35037 - Microsoft 365 audit logging is enabled
 Status: Pass
 Duration: 00:00:02.391
 Start: 2026-07-22 14:04:14.000
 End: 2026-07-22 14:04:16.392
 Version: 2.5.0
 ---
2026-07-22 14:04:13.995 [Verbose] Processing test '35037'
2026-07-22 14:04:14.001 [VeryVerbose] 🟦 Start
2026-07-22 14:04:14.003 [Verbose] Checking Purview audit logging configuration → Getting audit log configuration
2026-07-22 14:04:16.350 [Verbose] Retrieved audit log configuration
2026-07-22 14:04:16.389 [Verbose] Running tests → Purview audit logging enabled
2026-07-22 14:04:16.391 [Verbose] Adding test result detail for Purview audit logging enabled
2026-07-22 14:04:16.392 [Verbose] Processing test '35037' - Concluded

DeanGross144:
@AndrewSWX thanks, it looks like @astaykov removed this test, does this finding mean that it will be added back into the assessment?

PR #1437, opened 2026-07-22
Addresses #1434: retire check 35037

astaykov:
Addresses #1434 regarding check 35037

This check cannot be reliably performed in the context of Zero Trust Assessment framework and is being retired.

(maintainer reference for more details: https://github.com/microsoft/ztspecs/issues/78)


alexandair (approved):
LGTM

PR #1439, opened 2026-07-22
Fix cache check logic in Invoke-ZtGraphRequestCache

FriedrichWeinmann:
Updated to respect empty results from cache in most cases (some extremely rare concurrency issues might lead to an extra request, but should not be relevant in virtually any relevant case imaginable)

### DIFF
--- a/src/powershell/private/core/Invoke-ZtGraphRequestCache.ps1
+++ b/src/powershell/private/core/Invoke-ZtGraphRequestCache.ps1
@@ -82,10 +82,11 @@
 		$cacheKey = "$cacheKey|$headerString"
 	}
 	$cachedResult = $script:__ZtSession.GraphCache.Value[$cacheKey]
+	$isCached = $cachedResult -or $script:__ZtSession.GraphCache.Value.ContainsKey($CacheKey)
 	$isMethodGet = $Method -eq 'GET'
 
 	# Don't read from cache for batch requests.
-	if (-not $cacheBlocked -and -not $DisableCache -and -not $isBatch -and $cachedResult -and $isMethodGet) {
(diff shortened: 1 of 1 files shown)


src/powershell/private/core/Invoke-ZtGraphRequestCache.ps1
```diff
@@ -82,10 +82,11 @@
 		$cacheKey = "$cacheKey|$headerString"
 	}
 	$cachedResult = $script:__ZtSession.GraphCache.Value[$cacheKey]
```
alexandair, line 84: **Concurrent cache fill can return stale `$null`**

The value is read before checking key existence:

1. Runspace A reads a missing key → `$cachedResult = $null`.
2. Runspace B caches the Graph response.
3. Runspace A calls `ContainsKey()` → `$true`.
4. Runspace A treats this as a hit but returns its stale `$null`.

This produces missing assessment data rather than merely an extra Graph request. Since this cache is explicitly shared across parallel runspaces, check `ContainsKey()` first and retrieve the value only after confirming the hit. If the initial check misses and another runspace subsequently fills the cache, the worst outcome is the acknowledged duplicate request.

alexandair (changes requested):
@FriedrichWeinmann Please, address my feedback.

FriedrichWeinmann:
Heya @alexandair ,
you are absolutely correct - while both issues (pre-update or post-update) should be vanishingly rare, I agree that this scenario is the worse option.
I only considered the duplicate request and failed to notice the new problem this update causes.

I'll close this PR as undesirable.

Instead, this should be implemented upstream in the cache class by adding a `TryGetValue` method to retrieve that information thread-safely, then revisited on both `Invoke-ZtAzureRequestCache` and `Invoke-ZtGraphRequestCache` once available.
Upstream tracked here: https://github.com/PowershellFrameworkCollective/psframework/issues/739

PR #1457, opened 2026-07-28
Network - 27022 - Microsoft 365 traffic is protected end-to-end through Global Secure Access acquisition and compliant network enforcement

Manoj-Kesana:


tdetzner:
@alexandair can you pls have a look at the Posh code and review?

alexandair (changes requested):

## 4. Critical issues 🔴

### 4.1 Every parallel assessment is moved into the sequential dependency queue

The new split uses:

> `@($_.DependsOn).Count -gt 0`

at Invoke-ZtTests.ps1#L132-L135.

In PowerShell, `@($null).Count` is `1`, not `0`. Therefore, tests without dependencies also satisfy this predicate. The consequences are:

1. `$parallelTests` becomes empty.
2. `Start-ZtTestExecution` and `Wait-ZtTest` are bypassed.
3. All formerly parallel tests run sequentially through the loop at Invoke-ZtTests.ps1#L155-L157.
4. The overall `$Timeout` enforcement attached to `Wait-ZtTest` is also bypassed.

Additionally, `Get-ZtTestMetadata` does not define a `DependsOn` result property in its metadata object at Get-ZtTestMetadata.ps1#L157-L172. Named attributes are copied only when the destination property already exists, so `DependsOn` cannot currently be populated.

**Required correction:**

- Add `DependsOn = $null` to the metadata result object.
- Test dependency presence with an explicit null check, for example `$null -ne $_.DependsOn -and $_.DependsOn.Count -gt 0`.
- Add scheduler tests proving ordinary tests remain parallel and dependent tests run afterward.

### 4.2 The required composite circuit verdict is missing

The specification requires an explicit circuit result:

- Passed when both stages pass.
- Investigate when no stage fails but a stage requires investigation.
- Failed when either stage fails.
- Failure messaging must identify the open stage.

See the circuit determination requirements `/ztspecs/specs/network/27022.md#L112-L120`.

The generated tenant-info object only exposes the two raw statuses at Add-ZtOverviewM365ProtectionCircuit.ps1#L61-L68:

> `acquisitionStatus = $acquisitionStatus`  
> `enforcementStatus = $enforcementStatus`

It does not calculate `overallStatus`, identify open stages, explain the directional dependency when acquisition fails, or generate the required stage-specific message. The description at Add-ZtOverviewM365ProtectionCircuit.ps1#L55-L60 only describes flow sizing.

This leaves the dashboard without the core evaluation promised by spec `27022`.

### 4.3 Operational errors are rendered as confirmed security failures

The status routing only recognizes `Passed`, `Investigate`, and `Skipped`:

- Enforcement routing: Add-ZtOverviewM365ProtectionCircuit.ps1#L32-L43
- Acquisition routing: Add-ZtOverviewM365ProtectionCircuit.ps1#L45-L52

Any other status—including `Error` or `Planned`—falls into the red failure path:

- Enforcement becomes **“Acquired but not enforced.”**
- Acquisition becomes **“Unprotected - not acquired.”**

For example, an exception while running `25379` produces `TestStatus = 'Error'`, but the visualization asserts that enforcement is absent rather than unavailable. This converts inability to assess into a definite tenant security failure.

`Error`, `Planned`, missing results, and other non-verdict states should be represented as unavailable/investigate, not red fail. Only `Failed` should produce the confirmed failure terminals.

## 5. Warnings and recommendations ⚠️

### 5.1 `Get-ZtTestResultDetail` can miss results because of key-type differences

The helper accepts a string and directly indexes the session dictionary at Get-ZtTestResultDetail.ps1#L6-L14.

`Add-ZtTestResultDetail` can store `$actualTestId` as an integer when the ID comes from `[ZtTest()]` metadata rather than an explicitly supplied string. PowerShell dictionaries distinguish integer key `25376` from string key `'25376'`.

Normalize keys consistently—preferably to `[string]`—when both storing and retrieving results.

This helper is currently unused after the `27022` assessment was removed, but it exposes an unreliable framework API.

### 5.2 No behavioral tests cover the new aggregation

The only test-file change creates and mocks `Add-ZtOverviewM365ProtectionCircuit` at Invoke-ZtTenantInfo.Tests.ps1#L91-L115. There is no assertion that it is called for `Network` or `All`, and no tests for:

- Passed/failed/investigate circuit combinations.
- `Error`, `Skipped`, or missing child results.
- Device-count and normalized Sankey sizing.
- Disabled profile behavior.
- Structured `Data` persistence.
- Dependency scheduling.

The PR history previously contained comprehensive tests for these cases, but they were deleted. The new status and scheduling behavior should not merge without replacement coverage.

### 5.3 The dependency framework does not enforce dependencies

Even after correcting the null predicate, `DependsOn` only delays tests already selected for execution. It does not:

- Include missing child tests when a composite test is selected explicitly.
- Verify that dependencies produced results.
- Topologically sort dependency chains.
- Handle dependent tests classified into `$syncTests`.

This is currently dead infrastructure because no assessment declares `DependsOn`. Either remove the unused dependency changes from this dashboard-only PR or implement and test complete dependency semantics.

** Suggestions**

- Reuse the established unavailable-status normalization pattern from `Add-ZtOverviewPrivateAccess`.
- Include `overallStatus`, per-stage normalized statuses, `degraded`, and an actionable description in the tenant-info result.
- Remove `Get-ZtTestResultDetail` and `DependsOn` if `27022` is intentionally dashboard-only; they add framework risk without serving the current implementation.
- Consider omitting zero-width Sankey links to avoid unnecessary or unstable rendering behavior.

alexandair (approved):
LGTM

@Manoj-Kesana You did a great job.

PR #1461, opened 2026-07-29
SecOps - 41052 - Application control (WDAC / App Control for Business) is enforced via Intune

Manoj-Kesana:


src/powershell/tests/Test-Assessment.41052.ps1
```diff
@@ -0,0 +1,179 @@
...
+            $assignmentCount = $assignments.Count
+            $policyResults += [pscustomobject]@{
+                Name = $policy.name; TemplateFamily = $policy.templateReference.templateFamily; Mode = 'Unknown'
+                AssignmentCount = $assignmentCount; LastModified = $policy.lastModifiedDateTime
+                Status = if ($assignmentCount -gt 0) { 'Investigate' } else { 'Fail' }
+            }
+            continue
+        }
+
+        # WDAC is in audit mode only when its Enabled:Audit Mode rule is present.
```
alexandair: **Enforcement is inferred without positive evidence**

The mode parser classifies **every nonempty settings response that does not contain an “audit” substring as `Enforce`**:

- An XML policy whose content cannot be decoded or inspected becomes `Enforce`.
- “Enter XML data” selected without usable XML can become `Enforce`, even though Microsoft documents that configuration as **Not configured**.
- An unfamiliar or changed Graph setting shape becomes `Enforce`.
- A supplemental/custom policy without an observable mode can independently produce a pass.

The specification itself notes that custom XML mode can require parsing the uploaded XML value. Microsoft also documents that enforcement requires removal of the `Enabled:Audit Mode` rule; absence of that phrase from serialized Graph JSON is not evidence that the rule is absent from the underlying policy.

This can produce a security-significant false pass. The parser should default to `Unknown` and return `Enforce` only after positive interpretation of the relevant setting:

- For built-in controls, inspect the exact nested choice value representing enabled versus audit-only.
- For custom XML, decode and parse the policy content and inspect rule option 3.
- If the value is unavailable, unsupported, malformed, or opaque, retain `Unknown`.

src/powershell/tests/Test-Assessment.41052.ps1
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            $assignmentCount = $assignments.Count
+            $policyResults += [pscustomobject]@{
+                Name = $policy.name; TemplateFamily = $policy.templateReference.templateFamily; Mode = 'Unknown'
+                AssignmentCount = $assignmentCount; LastModified = $policy.lastModifiedDateTime
+                Status = if ($assignmentCount -gt 0) { 'Investigate' } else { 'Fail' }
+            }
+            continue
+        }
```
alexandair, line 104: **Per-policy errors are silently discarded**

The assignment and settings catch blocks in Test-Assessment.41052.ps1:69-101 do not log unexpected errors. They append an ambiguous result and continue, leaving support packages without the exception or affected endpoint.

Add sanitized `Write-PSFMessage` warnings for both failures, including the policy name and error record.

src/powershell/tests/Test-Assessment.41052.ps1
```diff
@@ -0,0 +1,179 @@
...
+            if ($statusCode -in 401, 403, 404) {
+                Add-ZtTestResultDetail -SkippedBecause NotApplicable -Result 'Microsoft Graph returned HTTP 401, 403, or 404. Verify Intune licensing, user RBAC, and DeviceManagementConfiguration.Read.All consent.'
+                return
+            }
+
+            $policyResults += [pscustomobject]@{
+                Name = $policy.name; TemplateFamily = $policy.templateReference.templateFamily; Mode = 'Unknown'
+                AssignmentCount = $null; LastModified = $policy.lastModifiedDateTime; Status = 'Investigate'
+            }
+            continue
```
alexandair, line 82: **Assignment failure produces an inaccurate result message**

When assignment retrieval fails, the policy is immediately marked `Investigate` in Test-Assessment.41052.ps1:76-80. The overall message then says an **assigned** policy was returned in Test-Assessment.41052.ps1:128-130, even though assignment could not be established.

Track assignment state separately and describe the outcome as “assignment or enforcement mode could not be determined.”

src/powershell/tests/Test-Assessment.41052.ps1
```diff
@@ -0,0 +1,179 @@
...
+    }
+    else {
+        $testResultMarkdown = "❌ No App Control for Business policies were returned, no policies are assigned, or all assigned policies are audit-only.`n`n%TestResult%"
+    }
+
+    #endregion Assessment Logic
+
+    #region Report Generation
+
+    $portalUrl = 'https://intune.microsoft.com/#view/Microsoft_Intune_Workflows/SecurityManagementMenu/~/applicationcontrol'
```
alexandair: **Portal route differs from Microsoft’s documented route**

The link uses `~/applicationcontrol`. Current Microsoft documentation identifies the route as `~/appControl`.

The specification contains the same outdated route and should be corrected along with the implementation.

alexandair (changes requested):

Suggestion: Link each policy name to its Intune policy summary instead of only linking the table heading. The repository already has the necessary deep-link pattern.

src/powershell/tests/Test-Assessment.41052.ps1
Manoj-Kesana: @millenniumfalcone - can you please update the SPEC?

alexandair (approved):
LGTM

PR #1470, opened 2026-07-31
SecOps - 41210 - Microsoft Threat Intelligence feeds are integrated into Microsoft Sentinel

aahmed-spec:


src/powershell/tests/Test-Assessment.41210.ps1
```diff
@@ -0,0 +1,364 @@
...
+        if ($forbiddenWorkspaces.Count -gt 0) {
+            # Auth errors mean we cannot confirm whether those workspaces have Sentinel onboarded;
+            # we cannot rule out a passing workspace exists among the inaccessible ones.
+            $params = @{
+                TestId       = '41210'
+                Title        = 'Microsoft Threat Intelligence feeds are integrated into Microsoft Sentinel'
+                Status       = $false
+                Result       = '⚠️ One or more Log Analytics workspaces returned insufficient permissions when checking Sentinel onboarding state. No Sentinel-onboarded workspace was confirmed among accessible workspaces — the overall state cannot be determined. Ensure Microsoft Sentinel Reader is granted on all workspaces and re-run the assessment.'
+                CustomStatus = 'Investigate'
+            }
```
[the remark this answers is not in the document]
aahmed-spec: made changes

src/powershell/tests/Test-Assessment.41210.ps1
```diff
@@ -0,0 +1,364 @@
...
+            RowStatus         = $rowStatus
+        }
+    }
+    $workspaceResults = @($workspaceResults)
+
+    $passedItems          = @($workspaceResults | Where-Object { $_.RowStatus -eq 'Pass' })
+    $passNoConnectorItems = @($workspaceResults | Where-Object { $_.RowStatus -eq 'PassNoConnector' })
+    $investigateItems     = @($workspaceResults | Where-Object { $_.RowStatus -eq 'Investigate' })
+
+    $passed       = ($passedItems.Count + $passNoConnectorItems.Count) -gt 0
```
alexandair, line 279: **A passing workspace masks other workspaces that require investigation**

Test sets `$passed` when **any** workspace passes:

```powershell
$passed = ($passedItems.Count + $passNoConnectorItems.Count) -gt 0
```

However, `Investigate` is selected only when there is no passing workspace, except for onboarding/helper errors. For example:

- Workspace A has indicators and passes.
- Workspace B returns HTTP 500 from Q2, or has a connector with zero indicators.
- `$passed` becomes `$true`.
- `$investigateItems` is ignored.
- The overall assessment reports a clean Pass.

This contradicts the specification’s explicit rule that an unexpected Q1/Q2 error returns **Investigate**. Include `$investigateItems` in the first Investigate branch regardless of whether another workspace passes.

The specification should also clarify whether a definitive `Fail` workspace may be masked by another passing workspace. The current “any workspace passes” aggregation is introduced by the implementation rather than stated in the specification.

src/powershell/tests/Test-Assessment.41210.ps1
```diff
@@ -0,0 +1,364 @@
...
+        $tiMetricsPath = "$($workspace.WorkspaceId)/providers/Microsoft.SecurityInsights/threatIntelligence/main/metrics?api-version=2024-09-01"
+
+        try {
+            $rawMetrics = @(Invoke-ZtAzureRequest -Path $tiMetricsPath -ErrorAction Stop)
+            $metricsByWorkspace[$workspace.WorkspaceId]   = $rawMetrics
+            $metricsOkByWorkspace[$workspace.WorkspaceId] = $true
+        }
+        catch {
+            $httpStatus = Get-ZtHttpStatusCode -ErrorRecord $_
+            if ($httpStatus -eq 404) {
```
alexandair: **Every Q2 HTTP 404 is treated as the documented zero-indicator response**

Test checks only the status:

```powershell
if ($httpStatus -eq 404) {
    $metricsByWorkspace[...] = @()
    $metricsOkByWorkspace[...] = $true
}
```

The specification limits this exception to HTTP 404 with **“No metrics data available.”** A different 404, such as a missing resource or unexpected route response, must be **Investigate**. With no connector present, the current code turns such a response into a definitive Fail.

`Invoke-ZtAzureRequest` preserves the ARM error message in the thrown exception, so this branch can verify both status and the documented message.

src/powershell/tests/Test-Assessment.41210.ps1
```diff
@@ -0,0 +1,364 @@
...
+        $sourceMetricsStr     = $null
+        $threatTypeMetricsStr = $null
+        $lastUpdated          = $null
+
+        if ($metricsOk) {
+            $indicatorCount       = 0
+            $sourceMetricsAgg     = @{}
+            $threatTypeMetricsAgg = @{}
+
+            foreach ($metricsItem in $rawMetrics) {
```
alexandair, line 208: **Schema-incomplete successful metrics responses silently become zero indicators**

Test initializes the count to zero and iterates directly over `properties.sourceMetrics` without validating the response shape.

If ARM returns HTTP 200 with an item missing `properties` or `sourceMetrics`, PowerShell performs no iterations and leaves `$indicatorCount = 0`. The result then becomes Fail or “connector present but no indicators,” rather than the specification-required **Investigate for an unexpected response shape**.

Validate that Q2 contains the documented metric structure and numeric `metricValue` entries before setting `$metricsOkByWorkspace` to `$true`.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.41210.ps1
aahmed-spec: made changes

src/powershell/tests/Test-Assessment.41210.ps1
aahmed-spec: made changes

src/powershell/tests/Test-Assessment.41210.ps1
aahmed-spec: made changes

src/powershell/tests/Test-Assessment.41210.ps1
```diff
@@ -143,13 +143,22 @@ function Test-Assessment-41210 {
 
         try {
             $rawMetrics = @(Invoke-ZtAzureRequest -Path $tiMetricsPath -ErrorAction Stop)
-            $metricsByWorkspace[$workspace.WorkspaceId]   = $rawMetrics
-            $metricsOkByWorkspace[$workspace.WorkspaceId] = $true
+            # A non-empty response must contain properties.sourceMetrics; absence is an unexpected shape.
+            $schemaOk = ($rawMetrics.Count -eq 0) -or ($null -ne $rawMetrics[0].properties -and $null -ne $rawMetrics[0].properties.sourceMetrics)
```
alexandair: **The new schema guard validates only the first metrics item**

```powershell
$schemaOk = ($rawMetrics.Count -eq 0) -or ($null -ne $rawMetrics[0].properties -and $null -ne $rawMetrics[0].properties.sourceMetrics)
```

The ARM contract returns `ThreatIntelligenceMetricsList.value` as an array. If the first item has `properties.sourceMetrics` but a later item does not, `$schemaOk` remains `$true`. Assessment logic then silently skips the missing collection and can emit a definitive Pass or Fail instead of the specification-required `Investigate` for an unexpected response shape.

A focused probe with one valid item followed by one malformed item produced:

```text
SchemaAccepted : True
IndicatorCount : 2
MalformedItems : 1
```

Validate every item before setting `$metricsOkByWorkspace[...] = $true`, for example by rejecting any item whose `properties` or `properties.sourceMetrics` is null. The same validation should occur before metric aggregation so no malformed element can be silently ignored.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.41210.ps1
aahmed-spec: made changes

alexandair (approved):
LGTM

PR #1481, opened 2026-08-04
SecOps - 41080 - Conditional Access App Control session policies are enforced for sensitive cloud apps

Manoj-Kesana:


src/powershell/tests/Test-Assessment.41080.ps1
```diff
@@ -0,0 +1,178 @@
...
+    #endregion Data Collection
+
+    #region Assessment Logic
+    $enabledPolicies = @($enabledPolicies)
+
+    $policyResults = foreach ($policy in $enabledPolicies) {
+        $targetApps = @($policy.conditions.applications.includeApplications | Where-Object { -not [string]::IsNullOrWhiteSpace($_) })
+        $sessionControl = $policy.sessionControls.cloudAppSecurity
+        $isEnabled = $null -ne $sessionControl -and $sessionControl.isEnabled -eq $true
+        $isMatchingPolicy = $isEnabled -and $targetApps.Count -gt 0 -and -not [string]::IsNullOrWhiteSpace($sessionControl.cloudAppSecurityType)
```
alexandair: **A policy satisfying the specification can be reported as failing**

The specification defines a match exclusively as an enabled cloud-app-security control targeting one or more applications:

> `sessionControls.cloudAppSecurity.isEnabled == true` and at least one targeted application.

See the pass criteria `/ztspecs/specs/secops/41080.md#L73` and evaluation logic `/ztspecs/specs/secops/41080.md#L81-L87`.

The implementation adds an undocumented requirement:

`$isMatchingPolicy = $isEnabled -and $targetApps.Count -gt 0 -and -not [string]::IsNullOrWhiteSpace($sessionControl.cloudAppSecurityType)`

Microsoft Graph defines `cloudAppSecurityType` as nullable. Therefore, a policy with:

- `isEnabled = true`
- one or more included applications
- a missing or null `cloudAppSecurityType`

is classified as nonmatching and can cause a definite **Fail**, even though it satisfies the specification's pass predicate.

This condition was introduced by the latest fix commit. A missing type should not be treated as proof that the session control is disabled. Either:

1. Follow the specification exactly and determine matching status from `isEnabled` plus targeted applications; or
2. Treat a missing type as `Investigate` if the type is considered necessary for reliable evaluation.

It should not become a definite failure.

src/powershell/tests/Test-Assessment.41080.ps1
```diff
@@ -0,0 +1,178 @@
...
+    else {
+        $testResultMarkdown = "❌ No enabled Conditional Access policy has Defender for Cloud Apps session control enabled for a cloud app.`n`n%TestResult%"
+    }
+    #endregion Assessment Logic
+
+    #region Report Generation
+    $portalUrl = 'https://entra.microsoft.com/#view/Microsoft_AAD_ConditionalAccess/ConditionalAccessBlade'
+    $policyUrlTemplate = 'https://entra.microsoft.com/#view/Microsoft_AAD_ConditionalAccess/PolicyBlade/policyId/{0}'
+    $maxDisplay = 10
+    $displayPolicies = @($policyResults | Sort-Object -Property @{ Expression = { -not $_.Matches } }, PolicyDisplayName | Select-Object -First $maxDisplay)
```
alexandair: **The output includes nonmatching policies contrary to the specification**

The specification says to list matching policies when the assessment passes. The implementation instead selects from every enabled policy at Test-Assessment.41080.ps1#L130:

`$displayPolicies = @($policyResults | Sort-Object -Property @{ Expression = { -not $_.Matches } }, PolicyDisplayName | Select-Object -First $maxDisplay)`

Consequently, an overall passing report can contain unrelated policies marked `❌ Fail`. This makes the table appear internally inconsistent and adds noise unrelated to the finding.

Use `$matchingPolicies` for pass and investigate output. For the no-match result, emit a clear empty-state row rather than listing arbitrary enabled Conditional Access policies.

alexandair (changes requested):

alexandair (approved):
LGTM

PR #1504, opened 2026-08-12
SecOps - 41055 - Real-time, behavioral, and heuristic protection are enabled on Microsoft Defender Antivirus

praneeth-0000:


src/powershell/tests/Test-Assessment.41055.ps1
```diff
@@ -0,0 +1,265 @@
...
+.NOTES
+    Test ID: 41055
+    Workshop Task ID: SECOPS-055
+    Category: Endpoint threat protection
+    Pillar: SecOps
+    Required Module: Microsoft.Graph.Authentication
+    Required Connection: Microsoft Graph
+#>
+
+function Test-Assessment-41055 {
```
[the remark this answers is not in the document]
praneeth-0000: Not needed

src/powershell/tests/Test-Assessment.41055.ps1
```diff
@@ -0,0 +1,269 @@
...
+        $testResultMarkdown = "❌ One or more of real-time, behavioral, or heuristic scanning is disabled.`n`n%TestResult%"
+    }
+    elseif ($passedItems.Count -gt 0) {
+        $passed = $true
+        $testResultMarkdown = "✅ Real-time, behavioral, and heuristic protection on Microsoft Defender Antivirus are enabled.`n`n%TestResult%"
+    }
+    else {
+        $customStatus = 'Investigate'
+        $testResultMarkdown = "⚠️ No Microsoft Defender Antivirus Secure Score control could be evaluated; verify SecurityEvents.Read.All is granted, Secure Score data is flowing, and at least one MDE device is onboarded.`n`n%TestResult%"
+    }
```
alexandair, line 216: **Row-level Investigate results can incorrectly roll up to Pass**

```powershell
$failedItems = @($evaluationResults | Where-Object Status -eq 'Fail')
$passedItems = @($evaluationResults | Where-Object Status -eq 'Pass')

if ($failedItems.Count -gt 0) {
    ...
}
elseif ($passedItems.Count -gt 0) {
    $passed = $true
    ...
}
else {
    $customStatus = 'Investigate'
}
```

The code creates an `Investigate` row when a returned score or profile maximum is null or nonnumeric at lines 167-169, but the roll-up never counts those rows. If one control passes and another is `Investigate`, `$passedItems.Count -gt 0` is true and the overall assessment reports Pass.

This contradicts the specification's explicit **Fail > Investigate > Pass** precedence at `/ztspecs/specs/secops/41055.md#L101`. It can certify the tenant while one pinned control could not be evaluated.

**Required change:** collect `Investigate` rows and apply the specified precedence: Fail when any row fails; otherwise Investigate when any row is unevaluable or when the evaluable set is empty; otherwise Pass.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.41055.ps1
```diff
@@ -0,0 +1,269 @@
...
+        $params = @{
+            TestId       = '41055'
+            Title        = 'Real-time, behavioral, and heuristic protection are enabled on Microsoft Defender Antivirus'
+            Status       = $false
+            Result       = '⚠️ No Microsoft Defender Antivirus Secure Score control could be evaluated; verify SecurityEvents.Read.All is granted, Secure Score data is flowing, and at least one MDE device is onboarded.'
+            CustomStatus = 'Investigate'
+        }
+        Add-ZtTestResultDetail @params
+        return
+    }
```
alexandair: **Empty Q1 catalog returns before listing the seven pinned controls**

Test immediately adds an Investigate result and returns when `$controlProfileById.Count -eq 0`.

That outcome is correct, but the early return omits the report table entirely. The current specification requires an empty evaluable set to list the N/A/ignored IDs at `/ztspecs/specs/secops/41055.md#L101`, and separately requires every pinned control to be listed at `/ztspecs/specs/secops/41055.md#L118`. Because all seven IDs are fixed and known, an empty Q1 response should render seven `N/A (profile not found)` rows before returning Investigate.

**Recommended change:** reserve the early return for actual query errors or a missing Secure Score snapshot. Let an empty Q1 catalog flow through row construction and the normal empty-evaluable-set roll-up so the report identifies all seven missing profiles.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.41055.ps1
praneeth-0000: Microsoft Graph documents score as Double and does not document null as an expected value, but as a defensive check I've implemented this. Thanks.

src/powershell/tests/Test-Assessment.41055.ps1
praneeth-0000: fixed

src/powershell/tests/Test-Assessment.41055.ps1
alexandair: **Empty Catalog Is Reported As Seven Controls Evaluated**

```powershell
$totalCount = $evaluationResults.Count
$countLine = "Total Microsoft Defender Antivirus controls evaluated: $totalCount"
```

Removing the `$controlProfileById.Count -eq 0` early return makes this pre-existing line reachable when Q1 returns no profiles. The implementation creates one result for each of the seven pinned IDs, so the report says:

```text
Total Microsoft Defender Antivirus controls evaluated: 7
```

In the same result, every row is `N/A (profile not found)` and the headline says no control could be evaluated. The count therefore describes the number of listed pinned controls, not the number evaluated, and contradicts the assessment result.

**Impact:** The overall `Investigate` status remains correct, but the user-facing evidence is internally inconsistent for the exact empty-catalog scenario added by this commit.

**Recommended correction:** Either label the value accurately, for example `Total pinned Microsoft Defender Antivirus controls: 7`, or calculate the evaluated count from statuses that actually participated in the roll-up.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.41055.ps1
praneeth-0000: I've renamed the label, thank you.

alexandair (approved):
LGTM

PR #1508, opened 2026-08-13
 Network - 27028 - A web content filtering policy governs Copilot Studio agent traffic through the baseline profile

Manoj-Kesana:


src/powershell/tests/Test-Assessment.27028.ps1
```diff
@@ -0,0 +1,151 @@
...
+    $testResultMarkdown = ''
+    $baselineState = 'Not found'
+    $enabledPolicyNames = @()
+
+    if ($httpStatusCode -eq 404) {
+        # Global Secure Access isn't provisioned in this tenant.
+        Write-PSFMessage 'Global Secure Access filtering profiles are not available in this tenant.' -Tag Test -Level Verbose
+        Add-ZtTestResultDetail -SkippedBecause NotApplicable
+        return
+    }
```
alexandair, line 76: **A missing baseline profile can be skipped instead of failed**

The implementation converts an HTTP 404 into `NotApplicable`:

```powershell
if ($httpStatusCode -eq 404) {
    Write-PSFMessage 'Global Secure Access filtering profiles are not available in this tenant.' -Tag Test -Level Verbose
    return
}
```

The controlling specification requires the opposite outcome: if Q1 returns no baseline profile, the test **fails**. See `/ztspecs/specs/network/27028.md#L83-L87`:

> If `value` is empty, or the returned baseline profile's `state` doesn't equal `enabled`, the test **Fails**.

The 404 branch explicitly interprets the condition as Global Secure Access not being provisioned. That is still absence of the required enforcement path, not evidence that the recommendation is inapplicable. Returning `NotApplicable` removes this high-risk finding from the tenant's failed controls and can make an unconfigured tenant appear healthier than it is.

**Required change:** treat a 404 representing an unavailable/unprovisioned filtering-profile resource as a normal failed assessment with baseline state `Not found`, consistent with the specified empty-result behavior. Reserve `Investigate` for indeterminate API failures and use `NotApplicable` only if the specification is first changed to define a genuine applicability condition.

src/powershell/tests/Test-Assessment.27028.ps1
```diff
@@ -0,0 +1,151 @@
...
+
+    if ($errorMsg) {
+        $customStatus = 'Investigate'
+        $testResultMarkdown = if ($httpStatusCode -in 401, 403) {
+            '⚠️ Unable to read the Global Secure Access baseline profile due to insufficient permissions. Grant the **NetworkAccess.Read.All** Microsoft Graph permission and assign the **Global Secure Access Administrator** or **Security Reader** role, then rerun the assessment.'
+        }
+        else {
+            '⚠️ Unable to retrieve the Global Secure Access filtering profiles due to an API error. Please rerun the assessment.'
+        }
+    }
```
alexandair: **The insufficient-permission remediation can be incomplete for the expanded query**

The 401/403 message tells the operator to grant only `NetworkAccess.Read.All`.

Microsoft documents `NetworkAccess.Read.All` for listing `filteringProfiles`, but documents `NetworkAccessPolicy.Read.All` for listing the linked `filteringPolicyLink` objects. This request expands `policies` and nested `policy`, and the specification also identifies `NetworkAccessPolicy.Read.All` at `/ztspecs/specs/network/27028.md#L108-L111`. If profile access succeeds but linked-policy access is denied, the current remediation can send the operator toward a permission they already have without naming the policy permission needed by the evaluated data.

**Recommendation:** make the message reflect all permissions required by the module's actual delegated request, or avoid naming one permission when the caught 403 does not identify which permission is missing.

src/powershell/tests/Test-Assessment.27028.ps1
```diff
@@ -0,0 +1,151 @@
...
+    # Q1: Get the baseline profile and its linked policies
+    $filteringProfiles = @()
+    $errorMsg = $null
+    $httpStatusCode = $null
+
+    try {
+        $filteringProfiles = Invoke-ZtGraphRequest -RelativeUri 'networkAccess/filteringProfiles' -QueryParameters @{
+            '$select' = 'id,name,state,priority'
+            '$expand' = 'policies($select=id,state;$expand=policy($select=id,name,version))'
+        } -ApiVersion beta -ErrorAction Stop
```
alexandair, line 57: **Q1 omits the specified server-side baseline filter**

The specification queries only `priority eq 65000` at `/ztspecs/specs/network/27028.md#L63-L68`. The implementation sends `$select` and `$expand` but no `$filter`, then locates the baseline client-side at `Test-Assessment.27028.ps1#L53-L57` and `Test-Assessment.27028.ps1#L89`.

Paging makes the current logic functionally correct, but it retrieves and expands every filtering profile when only one is relevant. Add `'$filter' = 'priority eq 65000'` to match Q1 and reduce Graph response size, cache size, and processing.

alexandair (changes requested):

Suggestion: Consider sorting with `-Unique` when rendering policy names. Duplicate names do not affect compliance, but deduplication would make the report more concise if multiple enabled links resolve to the same display name.

src/powershell/tests/Test-Assessment.27028.ps1
Manoj-Kesana: this is notapplicable cause the test makes a single GET /beta/networkAccess/filteringProfiles?...&$expand=policies(...) - that's the List filteringProfiles API (NetworkAccess.Read.All, delegated + application). It never calls /filteringProfiles/{id}/policies, which is the endpoint documented with NetworkAccessPolicy.Read.All. That permission is also delegated-only ("Application: Not supported"), so it isn't in Get-ZtGraphScope and can't be granted for app-only runs. The expanded policies/policy data returns fine under NetworkAccess.Read.All alone - same pattern as 25408, 25410, 25411, 25413, 25415, 25416, and 27000, and verified on a live tenant. Keeping the message as-is.

alexandair (approved):
LGTM

PR #1516, opened 2026-08-17
SecOps - 41201 - At least one Microsoft (first-party) data connector is configured in every Microsoft Sentinel workspace

Manoj-Kesana:


src/powershell/tests/Test-Assessment.41201.ps1
```diff
@@ -0,0 +1,274 @@
...
+        $customStatus       = 'Investigate'
+        $testResultMarkdown = "⚠️ The data-connectors API returned an unexpected response for one or more workspaces, or one or more workspaces could not be checked due to insufficient permissions. Note that connectors configured exclusively via diagnostic settings are not returned by this API and require a separate check. Re-run after verifying Microsoft Sentinel Reader access on each affected workspace.`n`n%TestResult%"
+    }
+    elseif ($passed) {
+        $testResultMarkdown = "✅ At least one Microsoft first-party data connector is configured in the Sentinel workspace.`n`n%TestResult%"
+    }
+    else {
+        $testResultMarkdown = "❌ No Microsoft first-party data connectors are configured in the Sentinel workspace.`n`n%TestResult%"
+    }
+
```
alexandair, line 214: **Investigate Incorrectly Overrides A Conclusive Failure**

```powershell
$investigateItems = @($workspaceResults | Where-Object { $_.RowStatus -eq 'Investigate' })
$failedItems      = @($workspaceResults | Where-Object { $_.RowStatus -eq 'Fail' })

$passed = $failedItems.Count -eq 0 -and $investigateItems.Count -eq 0 -and
          $forbiddenWorkspaces.Count -eq 0 -and $unresolvedWorkspaces.Count -eq 0

if ($investigateItems.Count -gt 0 -or $forbiddenWorkspaces.Count -gt 0 -or $unresolvedWorkspaces.Count -gt 0) {
    $customStatus = 'Investigate'
}
elseif ($passed) {
    # Pass
}
else {
    # Fail
}
```

The result precedence is wrong for an assessment whose requirement applies to **every** Sentinel workspace.

Consider two confirmed Sentinel workspaces:

| Workspace | Q1 result | Per-workspace status |
|---|---|---|
| A | Empty connector list | `Fail` |
| B | API error | `Investigate` |

Workspace A conclusively disproves the requirement. No possible result for workspace B can make “at least one first-party connector exists in every workspace” true. The overall result must therefore be `Fail`.

The current branch instead enters the first condition because `$investigateItems.Count -gt 0`, adds `CustomStatus = 'Investigate'`, and suppresses the definite failure. The same defect occurs when a known failed workspace is combined with a workspace whose onboarding check returned a permission or transient error.

This conflicts with the specification's definite failure rules for an empty connector list or a list containing only non-Microsoft kinds: `/ztspecs/specs/secops/41201.md#L75-L76`, `/ztspecs/specs/secops/41201.md#L96-L98`.

**Impact:** A High-risk, conclusively noncompliant tenant can be classified as ambiguous instead of failed. This weakens remediation prioritization and makes the assessment status depend incorrectly on whether an unrelated workspace also encountered an error.

**Required correction:** Apply result precedence in this order:

1. If any confirmed workspace has `RowStatus = 'Fail'`, return `Fail` without `CustomStatus`.
2. Otherwise, if any connector request, onboarding check, or permission check is unresolved, return `Investigate`.
3. Otherwise, return `Pass`.

This is the standard roll-up rule: `Fail > Investigate > Pass`.

src/powershell/tests/Test-Assessment.41201.ps1
```diff
@@ -0,0 +1,274 @@
...
+        return
+    }
+
+    if ($allWorkspaces -eq 'NoWorkspaces') {
+        Write-PSFMessage 'No Log Analytics workspaces found across accessible subscriptions — skipping Sentinel data connectors check.' -Tag Test -Level VeryVerbose
+        Add-ZtTestResultDetail -SkippedBecause NotApplicable
+        return
+    }
+
+    $checkableWorkspaces  = @($allWorkspaces | Where-Object { -not $_.PermissionError })
```
alexandair, line 79: **Workspaces Causing An Onboarding Investigate Result Are Missing From The Evidence Table**

**Locations:**

- Error collections: `Test-Assessment.41201.ps1#L81-L85`
- Result construction: `Test-Assessment.41201.ps1#L155-L185`
- Table rendering: `Test-Assessment.41201.ps1#L229-L249`

`$workspaceResults` is created only from `$onboardedWorkspaces`. Workspaces in `$forbiddenWorkspaces` or `$unresolvedWorkspaces` can force the overall result to `Investigate`, but they are never added to the report model.

For example, if one confirmed workspace passes and another workspace's Sentinel onboarding request returns 403, the report headline says one or more workspaces could not be checked, while the table shows only the passing workspace. The user cannot identify the affected subscription or workspace from the evidence supplied by this test.

**Recommended correction:** Add placeholder result rows for permission and onboarding errors with `RowStatus = 'Investigate'`, zero or unavailable connector counts, and an explicit reason. The neighboring assessment 41203 already uses this pattern through `$unresolvedWorkspaceResults`.

src/powershell/tests/Test-Assessment.41201.ps1
```diff
@@ -0,0 +1,274 @@
...
+                if ($null -ne $response -and $response.PSObject.Properties['value']) {
+                    $workspaceConnectors += @($response.value)
+                    $nextPath = $response.nextLink
+                }
+                else {
+                    $workspaceConnectors += @($response)
+                    $nextPath = $null
+                }
+            } while ($nextPath)
+
```
alexandair: **Remove The Redundant Manual Pagination Loop**

`Invoke-ZtAzureRequest` already enables ARM pagination for GET requests and returns the unwrapped item collection. The loop at `Test-Assessment.41201.ps1#L115-L129` therefore receives connector items rather than a response envelope in normal use and exits through its fallback branch.

This is not currently dropping records, but a direct assignment is clearer and matches neighboring Sentinel assessments:

```powershell
$connectorsByWorkspace[$workspace.WorkspaceId] = @(
    Invoke-ZtAzureRequest -Path $dataConnectorsPath -ErrorAction Stop
)
```

alexandair (changes requested):

alexandair (approved):
LGTM

PR #1523, opened 2026-08-19
SecOps - 41202 - Third-party (non-Microsoft) data connectors are configured in Microsoft Sentinel for non-Microsoft workloads in scope

aahmed-spec:


src/powershell/tests/Test-Assessment.41202.ps1
```diff
@@ -0,0 +1,388 @@
...
+'@
+
+    $connectorTableRows   = ''
+    $workspaceSummaryRows = ''
+    $maxDisplay           = 10
+    $summaryRowCount      = 0
+    $statusPriority       = @{ Investigate = 0; Pass = 1 }
+    $sortedResults        = @($workspaceResults | Sort-Object { $statusPriority[$_.RowStatus] }, SubscriptionName, WorkspaceName)
+
+    foreach ($wsResult in $sortedResults) {
```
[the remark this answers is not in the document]
aahmed-spec: made changes

src/powershell/tests/Test-Assessment.41202.ps1
```diff
@@ -0,0 +1,388 @@
...
+                    }
+                    elseif ($customerPublisher -and $publisher -eq $customerPublisher) {
+                        # Customer-authored codeless connector — out of scope (evaluated by spec 41203).
+                        $classification = 'Custom (41203)'
+                        $connStatus     = 'n/a'
+                    }
+                    elseif ($microsoftPublishers -contains $publisher) {
+                        $classification = 'Microsoft first-party'
+                        $connStatus     = 'n/a'
+                    }
```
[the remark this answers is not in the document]
aahmed-spec: not required

src/powershell/tests/Test-Assessment.41202.ps1
```diff
@@ -0,0 +1,388 @@
...
+                $kind           = $connector.kind
+                $publisher      = $null
+                $classification = 'Unclassifiable'
+                $connStatus     = 'Investigate'
+
+                if ($kind -eq 'StaticUI') {
+                    # Content-hub connector — first-party, publisher not surfaced by API.
+                    $publisher      = '—'
+                    $classification = 'Microsoft first-party'
+                    $connStatus     = 'n/a'
```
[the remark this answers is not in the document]
aahmed-spec: not required

src/powershell/tests/Test-Assessment.41202.ps1
```diff
@@ -0,0 +1,388 @@
...
+    Workshop Task: SECOPS_094
+    Pillar: SecOps
+    Category: Security information and event management
+    Required API: Azure Resource Manager (management.azure.com)
+#>
+function Test-Assessment-41202 {
+    [ZtTest(
+        Category = 'Security information and event management',
+        ImplementationCost = 'Medium',
+        MinimumLicense = ('Consumption-based: Microsoft Sentinel'),
```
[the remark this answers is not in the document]
aahmed-spec: not required

src/powershell/tests/Test-Assessment.41202.ps1
```diff
@@ -0,0 +1,400 @@
...
+
+    Write-PSFMessage '🟦 Start' -Tag Test -Level VeryVerbose
+    $activity = 'Checking third-party data connectors in Microsoft Sentinel workspaces'
+    $azContext = Get-AzContext -ErrorAction SilentlyContinue
+    $tenantId = if ($azContext) { [string]$azContext.Tenant.Id } else { $null }
+    # Identify the customer's publisher name so customer-authored codeless connectors (scope of 41203) are excluded from third-party classification.
+    $customerPublisher = if ($tenantId) { Get-ZtTenantName -TenantId $tenantId } else { $null }
+    if ($customerPublisher -eq $tenantId) {
+        $customerPublisher = $null
+    }
```
alexandair: **Failure to resolve the customer publisher can produce a false Pass**

The code resolves the tenant display name and deliberately converts the helper's tenant-ID fallback to `$null` at `Test-Assessment.41202.ps1#L36-L42`:

```powershell
$customerPublisher = if ($tenantId) { Get-ZtTenantName -TenantId $tenantId } else { $null }
if ($customerPublisher -eq $tenantId) {
    $customerPublisher = $null
}
```

That fallback is reachable because `/zerotrustassessment/src/powershell/private/core/Get-ZtTenant.ps1#L20-L26` silently returns an empty string when its Graph lookup fails, after which `/zerotrustassessment/src/powershell/private/core/Get-ZtTenantName.ps1#L21-L28` returns the tenant ID.

The classifier only checks for a customer-authored connector when `$customerPublisher` is truthy. If it is null, every populated publisher other than Microsoft enters the final third-party branch `Test-Assessment.41202.ps1#L212-L225`. A customer-authored codeless connector can therefore set `$thirdPartyFound = $true` and make the workspace Pass even though the specification assigns that connector to 41203 and explicitly excludes it from 41202.

**Required change:** make successful customer-identity resolution a prerequisite for publisher-based third-party classification. If it cannot be resolved, classify non-Microsoft codeless publishers as Unclassifiable/Investigate, or use a reliable tenant-owned publisher source. Dedicated AWS/GCP kinds can still be classified independently.

src/powershell/tests/Test-Assessment.41202.ps1
```diff
@@ -0,0 +1,400 @@
...
+        Write-PSFMessage 'No enabled subscriptions found — skipping Sentinel third-party data connectors check.' -Tag Test -Level VeryVerbose
+        Add-ZtTestResultDetail -SkippedBecause NotApplicable
+        return
+    }
+
+    if ($allWorkspaces -eq 'NoWorkspaces') {
+        Write-PSFMessage 'No Log Analytics workspaces found across accessible subscriptions — skipping Sentinel third-party data connectors check.' -Tag Test -Level VeryVerbose
+        Add-ZtTestResultDetail -SkippedBecause NotApplicable
+        return
+    }
```
alexandair, line 85: **No subscriptions or workspaces returns Skipped instead of Investigate**

The two sentinel values from `Get-SentinelWorkspaceData` are converted to `SkippedBecause NotApplicable` at `Test-Assessment.41202.ps1#L76-L85`:

```powershell
if ($allWorkspaces -eq 'NoSubscriptions') {
    return
}

if ($allWorkspaces -eq 'NoWorkspaces') {
    return
}
```

The specification requires `Investigate` whenever no Sentinel-onboarded workspace can be evaluated and states that Skipped is reserved for the engine's license gate at `/ztspecs/specs/secops/41202.md#L68-L74` and `/ztspecs/specs/secops/41202.md#L113-L115`. Both branches are cases where no onboarded workspace can be evaluated, so they currently suppress a required assessment result.

**Required change:** return `Status = $false`, `CustomStatus = 'Investigate'`, and a reason specific to no accessible subscriptions or no Log Analytics workspaces.

src/powershell/tests/Test-Assessment.41202.ps1
```diff
@@ -0,0 +1,400 @@
...
+                }
+            }
+        }
+
+        # Workspace summary row — capped at $maxDisplay to match peer pattern.
+        if ($summaryRowCount -lt $maxDisplay) {
+            $wsStatusDisplay      = if ($wsResult.RowStatus -eq 'Pass') { '✅ Pass' } else { '⚠️ Investigate' }
+            $workspaceSummaryRows += "| $subMd | $workspaceMd | $($wsResult.ThirdPartyCount) | $($wsResult.FirstPartyCount) | $($wsResult.CustomCount) | $($wsResult.UnclassifiableCount) | $wsStatusDisplay |`n"
+            $summaryRowCount++
+        }
```
alexandair: **Output truncation breaks the required traceability**

Report generation globally caps connector rows and workspace summary rows at ten at `Test-Assessment.41202.ps1#L335-L371`, replacing the omitted evidence with count text at `Test-Assessment.41202.ps1#L374-L381`.

The specification requires one row per Q1 connector and a per-workspace roll-up at `/ztspecs/specs/secops/41202.md#L124-L138`. More importantly, it requires every workspace status to be listed so an aggregate Investigate remains traceable at `/ztspecs/specs/secops/41202.md#L121-L122`. In a tenant with more than ten workspaces, the report can omit the workspace responsible for the aggregate verdict. The generic Sentinel portal link does not preserve the assessment-time classification, error, or counts.

**Required change:** render every workspace summary. Also render every connector as specified, or revise the specification to define an evidence-preserving truncation rule. At minimum, decisive Investigate rows must never be omitted.

src/powershell/tests/Test-Assessment.41202.ps1
```diff
@@ -0,0 +1,400 @@
...
+
+        try {
+            # Q2: definitions expose publisher for Codeless Connector Framework entries; fallback only.
+            $q2Path = "$($workspace.WorkspaceId)/providers/Microsoft.SecurityInsights/dataConnectorDefinitions?api-version=2024-09-01"
+            $definitionsByWorkspace[$workspace.WorkspaceId] = @(Invoke-ZtAzureRequest -Path $q2Path -ErrorAction Stop)
+        }
+        catch {
+            $definitionsByWorkspace[$workspace.WorkspaceId] = $null
+            Write-PSFMessage "Q2 error querying connector definitions for workspace '$($workspace.WorkspaceName)' in subscription '$($workspace.SubscriptionName)': $_" -Tag Test -Level Warning
+        }
```
alexandair: **Q2 is executed for every workspace instead of only as a fallback**

The implementation calls `dataConnectorDefinitions` unconditionally immediately after Q1 at `Test-Assessment.41202.ps1#L124-L131`. The specification defines Q2 as a fallback only when a codeless Q1 item has a null or empty publisher and says its additional permission is needed only in that case at `/ztspecs/specs/secops/41202.md#L84-L96` and `/ztspecs/specs/secops/41202.md#L102`.

This adds one ARM request per workspace, increases throttling exposure, and emits avoidable permission warnings. It also creates inconsistent error semantics: a Q2 failure can be logged while the workspace still Passes when no publisher fallback was needed.

**Recommendation:** inspect Q1 first and call Q2 lazily only for workspaces containing unresolved codeless publishers. If that required fallback fails or is malformed, mark the workspace Investigate.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.41202.ps1
aahmed-spec: made changes

src/powershell/tests/Test-Assessment.41202.ps1
aahmed-spec: made changes

src/powershell/tests/Test-Assessment.41202.ps1
aahmed-spec: The spec has been updated since your review. The spec author revised the spec to explicitly say: *"If every accessible subscription was enumerated successfully and none contains a Sentinel-onboarded workspace, return Skipped."* The current code uses `Add-ZtTestResultDetail -SkippedBecause NotApplicable` for both branches, which is now correct per the updated spec. The `onboardedWorkspaces.Count -eq 0` block still returns Investigate when `$forbiddenWorkspaces` or `$unresolvedWorkspaces` is non-empty (i.e., when a scope couldn't be read).

src/powershell/tests/Test-Assessment.41202.ps1
aahmed-spec: made changes

src/powershell/tests/Test-Assessment.41202.ps1
```diff
@@ -0,0 +1,393 @@
...
+    # Identify the customer's publisher name so customer-authored codeless connectors (scope of 41203) are excluded from third-party classification.
+    $customerPublisher = if ($tenantId) { Get-ZtTenantName -TenantId $tenantId } else { $null }
+
+    # D1 + D2: workspace discovery and Sentinel onboarding check via shared helper.
+    # Returns 'Forbidden'        on ARG 401/403 — Investigate.
+    # Returns $null              on unexpected ARG failure — Investigate.
+    # Returns 'NoSubscriptions'  when no enabled subscriptions accessible — Skipped.
+    # Returns 'NoWorkspaces'     when no Log Analytics workspaces found — Skipped.
+    # Returns array              per-workspace results; PermissionError/OnboardingError mark inaccessible workspaces.
+    $allWorkspaces = Get-SentinelWorkspaceData -Activity $activity
```
alexandair, line 49: **Aggregate ARG discovery cannot preserve per-subscription read failures**

The updated specification requires D1 to enumerate Log Analytics workspaces **in each accessible subscription**, and any unreadable scope must become Investigate at `ztspecs/specs/secops/41202.md#L62-L74` and `/ztspecs/specs/secops/41202.md#L116`.

The test delegates D1 to `Get-SentinelWorkspaceData`. That helper performs one tenant-scoped Resource Graph query at `/zerotrustassessment/src/powershell/private/tests-shared/Get-SentinelWorkspaceData.ps1#L57-L94`, rather than one scoped workspace-list request per subscription. `Invoke-ZtAzureResourceGraphRequest` explicitly runs across all accessible subscriptions when no subscription IDs are supplied at `/zerotrustassessment/src/powershell/public/Invoke-ZtAzureResourceGraphRequest.ps1#L28-L36`.

This aggregate query has no per-subscription result or error record. If the caller can enumerate a subscription but cannot read its workspaces, the implementation cannot represent that subscription as an unresolved scope; it can only consume the resources ARG returns. A partially visible tenant can therefore omit a Sentinel workspace and still Pass, or return `NoWorkspaces` and Skip.

**Required change:** implement D1 per accessible subscription using the specified ARM workspace-list endpoint, or issue independently scoped discovery requests that preserve an error result for every subscription. Roll any 401/403/429/5xx or malformed response into an Investigate workspace/scope record before evaluating whether a skip is allowed.

alexandair, line 49: **A malformed HTTP 200 onboarding response is accepted as onboarded**

The specification says D2 must return Investigate for a malformed body, even when the request itself succeeds, at `/ztspecs/specs/secops/41202.md#L69-L74` and `/ztspecs/specs/secops/41202.md#L116`.

The shared helper inspects only `StatusCode` and unconditionally sets `$sentinelOnboarded = $true` for every 200 response at `/zerotrustassessment/src/powershell/private/tests-shared/Get-SentinelWorkspaceData.ps1#L118-L150`:

```powershell
switch ([int]$response.StatusCode) {
    200 { $sentinelOnboarded = $true }
    404 { $sentinelOnboarded = $false }
    # ...
}
```

There is no parse or schema check of `$response.Content`. An empty, invalid JSON, or structurally invalid onboarding-state body is therefore treated as a confirmed Sentinel workspace. If Q1 then returns a classifiable third-party connector, the workspace and tenant can Pass despite the unresolved D2 signal.

**Required change:** for HTTP 200, parse `Content` and validate the expected onboarding-state resource before setting `SentinelOnboarded`. Set `OnboardingError = $true` for empty, invalid, or malformed bodies so the existing tenant roll-up produces Investigate.

src/powershell/tests/Test-Assessment.41202.ps1
```diff
@@ -0,0 +1,393 @@
...
+    $testTitle = 'Third-party (non-Microsoft) data connectors are configured in Microsoft Sentinel for non-Microsoft workloads in scope'
+
+    #region Data Collection
+
+    Write-PSFMessage '🟦 Start' -Tag Test -Level VeryVerbose
+    $activity = 'Checking third-party data connectors in Microsoft Sentinel workspaces'
+    $azContext = Get-AzContext -ErrorAction SilentlyContinue
+    $tenantId = if ($azContext) { [string]$azContext.Tenant.Id } else { $null }
+    # Identify the customer's publisher name so customer-authored codeless connectors (scope of 41203) are excluded from third-party classification.
+    $customerPublisher = if ($tenantId) { Get-ZtTenantName -TenantId $tenantId } else { $null }
```
alexandair, line 39: **Tenant-name lookup failure still allows a customer connector to Pass as third-party**

The test obtains `$customerPublisher` directly from `Get-ZtTenantName` `Test-Assessment.41202.ps1#L34-L39`. That helper does not return null on lookup failure; it returns the tenant GUID  `/zerotrustassessment/src/powershell/private/core/Get-ZtTenantName.ps1#L20-L27`, because `Get-ZtTenant` silently returns an empty string when its Graph lookup fails at `/zerotrustassessment/src/powershell/private/core/Get-ZtTenant.ps1#L20-L28`.

The classifier treats any truthy `$customerPublisher` as resolved. It only emits Unclassifiable when that value is false, then classifies every non-Microsoft publisher that differs from the tenant GUID as third-party at `Test-Assessment.41202.ps1#L205-L241`:

```powershell
elseif (-not $customerPublisher) {
    $classification = 'Unclassifiable'
    # ...
}
else {
    $classification = 'Third-party'
    $thirdPartyFound = $true
}
```

For a customer-authored connector whose free-form publisher is the organization name, a failed tenant lookup yields `customerPublisher = <tenant-guid>`. The names do not match, so the connector is counted as third-party and can produce a false Pass. Microsoft Learn documents `connectorUiConfig.publisher` as a free-form provider string, not a tenant identifier.

**Required change:** treat the tenant-ID fallback as unresolved before classification, for example by resetting `$customerPublisher` to `$null` when it equals `$tenantId`. More robustly, use a reliable organization-owned publisher source; when ownership cannot be established, codeless non-Microsoft publishers must remain Unclassifiable/Investigate. Add a regression test covering tenant-name lookup failure with a customer-authored connector.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.41202.ps1
aahmed-spec: Closing this comment as discussed, malformed is not required anymore.

aahmed-spec: Closing this comment as discussed, malformed is not required anymore.

src/powershell/tests/Test-Assessment.41202.ps1
aahmed-spec: made changes

alexandair (approved):
LGTM

PR #1527, opened 2026-08-20
Secops - 41205 - Data collection rules (DCRs) are used to control ingestion into Microsoft Sentinel workspaces

Manoj-Kesana:


src/powershell/tests/Test-Assessment.41205.ps1
```diff
@@ -0,0 +1,363 @@
...
+            # A confirmed match is proof per spec, so uncertainty about other rules cannot override it.
+            $rowStatus = if ($dcrCount -ge 1) {
+                'Pass'
+            }
+            elseif ($dcrState.HasDetailErrors -or $dcrState.HasUnresolvedDestinations) {
+                'Investigate'
+            }
+            else {
+                'Fail'
+            }
```
alexandair: **An unrelated unresolved destination converts a definite workspace failure to Investigate**

The implementation records one subscription-wide `HasUnresolvedDestinations` flag whenever any successfully fetched DCR references a workspace outside the ARG inventory at `Test-Assessment.41205.ps1#L139-L166`. It then applies that flag to every onboarded workspace in the subscription that has no matching DCR:

```powershell
$rowStatus = if ($dcrCount -ge 1) {
    'Pass'
}
elseif ($dcrState.HasDetailErrors -or $dcrState.HasUnresolvedDestinations) {
    'Investigate'
}
else {
    'Fail'
}
```

Affected roll-up: `Test-Assessment.41205.ps1#L216-L231`.

A successfully fetched destination with a different resource ID cannot target the current known workspace. For example, if workspace A is onboarded and no DCR targets A, while one fetched DCR targets inaccessible workspace B, the current code assigns A `Investigate`. The specification requires A to **Fail** because no `workspaceResourceId` matches A. Since tenant precedence treats any `Investigate` as the overall outcome when no row is marked Fail, this can suppress the definite failure entirely.

A focused reproduction produced `Matched DCRs: 0; current implementation status: Investigate; spec status for known unmatched workspace: Fail`.

**Required change:** do not use a subscription-wide unresolved-destination flag to classify a known workspace. Preserve `Investigate` for Q1/Q2 failures where an unread rule could still target the workspace, but classify a workspace `Fail` when all successfully retrieved destination IDs are known not to match it. If unresolved workspaces must remain visible, represent them as separate investigate rows rather than allowing them to override known workspace verdicts.

src/powershell/tests/Test-Assessment.41205.ps1
```diff
@@ -0,0 +1,363 @@
...
+    $tableRows      = ''
+    $maxDisplay     = 10
+    $statusPriority = @{ Fail = 0; Investigate = 1; Pass = 2 }
+    $displayResults = @($workspaceResults | Sort-Object { $statusPriority[$_.RowStatus] }, SubscriptionName, WorkspaceName)
+    $hasMoreItems   = $false
+    if ($workspaceResults.Count -gt $maxDisplay) {
+        $displayResults = @($displayResults | Select-Object -First $maxDisplay)
+        $hasMoreItems   = $true
+    }
+
```
alexandair: **The 10-item report cap limits workspaces, not rendered DCR rows**

The code selects up to ten workspace result objects, then emits one row for every matching DCR within each selected workspace at `Test-Assessment.41205.ps1#L322-L338`. One workspace with 25 matching DCRs therefore renders 25 rows. The “remaining” calculation also counts omitted workspaces rather than omitted table rows at `Test-Assessment.41205.ps1#L342-L345`.

A focused reproduction selected one workspace but rendered 25 DCR rows with `$maxDisplay = 10`.

**Recommendation:** flatten workspace/DCR display rows first, preserve at least one row per decisive workspace as needed, then apply a documented row cap and calculate the omitted count from the flattened rows. At minimum, rename the limit and message so they accurately describe a workspace cap.

alexandair (changes requested):
@Manoj-Kesana Please, address my feeback.

src/powershell/tests/Test-Assessment.41205.ps1
```diff
@@ -0,0 +1,390 @@
...
+            continue
+        }
+
+        $dcrNameMd     = if ([string]::IsNullOrWhiteSpace($dcr.Name)) { 'Unknown rule' } else { Get-SafeMarkdown $dcr.Name }
+        $dcrMd         = if ([string]::IsNullOrWhiteSpace($dcr.ResourceId)) { $dcrNameMd } else { "[$dcrNameMd]($portalHost/#resource$($dcr.ResourceId)/overview)" }
+        $destinationMd = if ($dcr.MatchingDestinationNames.Count -gt 0) { ($dcr.MatchingDestinationNames | ForEach-Object { Get-SafeMarkdown $_ }) -join ', ' } else { '—' }
+        $transformMd   = if (-not $dcr.RoutesToDestination) { '—' } elseif ($dcr.TransformPresent) { '✅ Yes' } else { '⚠️ No' }
+        $reasonMd      = Get-SafeMarkdown $dcr.Reason
+
+        $tableRows += "| $subMd | $workspaceMd | $dcrMd | $destinationMd | $routesMd | $sourcesMd | $transformMd | $($statusDisplay[$dcr.Status]) | $reasonMd | $rollupMd |`n"
```
alexandair, line 368: **Two Required Report Columns Are Always Blank**

**Location:** `Test-Assessment.41205.ps1#L358-L366`  
**Specification:** `/ztspecs/specs/secops/41205.md#L101-L107`

The table row uses `$routesMd` and `$sourcesMd`, but neither variable is assigned anywhere in the function:

```powershell
$destinationMd = if ($dcr.MatchingDestinationNames.Count -gt 0) { ... } else { '—' }
$transformMd   = if (-not $dcr.RoutesToDestination) { '—' } elseif ($dcr.TransformPresent) { '✅ Yes' } else { '⚠️ No' }
$reasonMd      = Get-SafeMarkdown $dcr.Reason

$tableRows += "| ... | $destinationMd | $routesMd | $sourcesMd | $transformMd | ... |"
```

Consequently, every evaluated DCR row omits both required values:

- `Data flow routes to destination`
- `Data source types`

A focused mocked pass-path execution produced:

```text
Status=True
| Sub One | Workspace One | dcr-one | la-one |  |  | ✅ Yes | ✅ Pass | Telemetry is routed with an ingestion transform. | ✅ Pass |
```

This is misleading because the row says `Pass` but does not show the routing information that explains why it passed. Assign both variables before adding the row, for example:

```powershell
$routesMd = if ($dcr.RoutesToDestination) { '✅ Yes' } else { '❌ No' }
$sourcesMd = if ($dcr.DataSourceTypes.Count -gt 0) {
    ($dcr.DataSourceTypes | ForEach-Object { Get-SafeMarkdown $_ }) -join ', '
}
else {
    '—'
}
```

alexandair (changes requested):

alexandair (approved):
LGTM

PR #1528, opened 2026-08-21
Secops - 41049 - The Microsoft Defender Antivirus user interface is hidden from end users

ashwinikarke:


src/powershell/tests/Test-Assessment.41049.ps1
```diff
@@ -0,0 +1,366 @@
...
+                }
+            }
+        }
+        catch {
+            Add-ZtTestResultDetail @investigateParams
+            return
+        }
+    }
+
+    # Q2: Legacy windows10EndpointProtectionConfiguration (fallback if no Settings Catalog results)
```
alexandair: **Context-only Q1 rows prevent the required Q2 fallback**

The specification requires Q2 when Q1 has no `DisableVirusUI` instance at `/ztspecs/specs/secops/41049.md#L150-L152. The implementation instead runs Q2 only when the entire result list is empty:

```powershell
# Q2: Legacy windows10EndpointProtectionConfiguration (fallback if no Settings Catalog results)
if ($evaluationResults.Count -eq 0) {
```

Affected condition: `Test-Assessment.41049.ps1#L171-L177`.

Q1 defines one primary and 14 contextual controls at `Test-Assessment.41049.ps1#L103-L120`, and every matched control is appended to the same list at Test-Assessment.41049.ps1#L152-L161`. A policy containing only `DisableAccountProtectionUI`, for example, makes the list nonempty and suppresses Q2 even though the primary control is absent.

**Reproduction:** Q1 returned one assigned Account UI row; Q2 was mocked with an assigned legacy `defenderSecurityCenterDisableVirusUI = true` profile. Expected: Pass. Actual: Investigate, with zero Q2 calls.

**Required change:** run Q2 when Q1 has no result whose `IsPrimary` is true, while retaining Q1 contextual rows for output.

src/powershell/tests/Test-Assessment.41049.ps1
```diff
@@ -0,0 +1,366 @@
...
+
+    #region Assessment Logic
+
+    # Q1 error is only fatal when Q2 also failed to recover any data.
+    $hasQueryError = ($evaluationResults.Count -eq 0) -and (($null -ne $settingsCatalogError) -or ($null -ne $legacyEppError))
+
+    if ($hasQueryError) {
+        Add-ZtTestResultDetail @investigateParams
+        return
+    }
```
alexandair, line 283: **A Q1 list failure can be masked by a passing Q2 result**

The Q1 collection catch records `$settingsCatalogError` at `Test-Assessment.41049.ps1#L59-L62`, but the later error gate only considers that error when no evaluation rows were recovered:

```powershell
$hasQueryError = ($evaluationResults.Count -eq 0) -and (($null -ne $settingsCatalogError) -or ($null -ne $legacyEppError))
```

Affected condition: `Test-Assessment.41049.ps1#L271-L276`.

If Q1 fails and Q2 returns an assigned enabled primary row, `$evaluationResults.Count` is nonzero, `$hasQueryError` is false, and the Pass branch at `Test-Assessment.41049.ps1#L290-L305` wins. The specification classifies Graph access errors as Investigate and defines `Fail > Investigate > Pass` precedence at `/ztspecs/specs/secops/41049.md#L118-L124` and `/ztspecs/specs/secops/41049.md#L153-L158`.

**Reproduction:** Q1 list threw an error; Q2 returned an assigned legacy `DisableVirusUI = true` profile. Expected: Investigate. Actual: Pass.

**Required change:** carry a Q1 query failure into the roll-up as Investigate evidence. A known Fail may still outrank it, but a Pass must not.

src/powershell/tests/Test-Assessment.41049.ps1
```diff
@@ -0,0 +1,366 @@
...
+            $customStatus = 'Investigate'
+            $testResultMarkdown = "⚠️ Could not determine `DisableVirusUI` state: no matching control on either surface, empty result set, uninterpretable value, or Graph 401/403/404.`n`n%TestResult%"
+        }
+        else {
+            $passResults = @($primaryResults | Where-Object { $_.Status -eq 'Pass' })
+            if ($passResults.Count -gt 0) {
+                $passed = $true
+                $testResultMarkdown = "✅ The Microsoft Defender Antivirus (Virus & threat protection) UI is hidden from end users — an assigned policy enables `DisableVirusUI`.`n`n%TestResult%"
+            }
+        }
```
alexandair, line 321: **Per-policy errors return before `Fail > Investigate > Pass` is applied**

Both detail loops immediately emit Investigate and return:

```powershell
catch {
    Add-ZtTestResultDetail @investigateParams
    return
}
```

Affected catches: Settings Catalog lines `Test-Assessment.41049.ps1#L165-L168` and legacy EPP lines `Test-Assessment.41049.ps1#L260-L263`.

These returns bypass the roll-up at `Test-Assessment.41049.ps1#L290-L307`. They can therefore replace an already-proven Fail with Investigate or stop before a later failing policy is evaluated, contrary to the specification's precedence at `/ztspecs/specs/secops/41049.md#L158`.

**Reproduction:** the first Q1 policy returned assigned `DisableVirusUI = _0`; reading the second policy threw. Expected: Fail. Actual: Investigate.

**Required change:** record per-policy read failures as Investigate evidence, log the error, continue processing usable policies, and perform one final precedence roll-up.

src/powershell/tests/Test-Assessment.41049.ps1
```diff
@@ -0,0 +1,366 @@
...
+                @($settings | Where-Object {
+                    $_.settingInstance.settingDefinitionId -match $Pattern
+                })
+            }
+
+            $assignmentCount = $assignments.Count
+
+            # Extract all UI suppression controls for context
+            $uiSuppressionControls = @(
+                @{ Name = 'Virus/AV UI'; Pattern = '(?i)_disablevirusui(?:_|$)'; IsPrimary = $true },
```
alexandair: **The primary setting-definition match is not exact**

The primary pattern is:

```powershell
'(?i)_disablevirusui(?:_|$)'
```

Affected definition: `Test-Assessment.41049.ps1#L105`. It accepts an exact suffix but also accepts longer IDs beginning `_disablevirusui_`. The specification requires an exact, case-insensitive setting-definition match at `/ztspecs/specs/secops/41049.md#L93-L98` and `/ztspecs/specs/secops/41049.md#L151`.

A synthetic `_disablevirusui_override` setting reproduced a false Pass. No current colliding Graph setting ID was identified, so this is a warning rather than a demonstrated tenant-data failure.

**Recommendation:** compare the full pinned definition ID case-insensitively, or at minimum anchor `_disablevirusui$`. Apply exact matching to the contextual IDs as well.

src/powershell/tests/Test-Assessment.41049.ps1
```diff
@@ -0,0 +1,366 @@
...
+                    if ($null -eq $instance) { continue }
+                    [void]$flattenedSettings.Add($instance)
+                    $childInstances = @($instance.settingInstance.choiceSettingValue.children | Where-Object { $_ })
+                    foreach ($gsv in @($instance.settingInstance.groupSettingCollectionValue | Where-Object { $_ })) {
+                        $childInstances += @($gsv.children | Where-Object { $_ })
+                    }
+                    $childInstances += @($instance.settingInstance.groupSettingValue.children | Where-Object { $_ })
+                    $wrapped = @($childInstances | ForEach-Object { [PSCustomObject]@{ settingInstance = $_ } })
+                    if ($wrapped.Count -gt 0) { & $visitSettingInstances $wrapped }
+                }
```
alexandair, line 86: **Recursive traversal omits choice-setting collection children**

The walker visits singular choice children and group children at `Test-Assessment.41049.ps1#L75-L88`, but it never visits `choiceSettingCollectionValue[].children`.

Microsoft Graph defines `choiceSettingCollectionValue` as a collection of choice values, and each [choice value can contain child setting instances](https://learn.microsoft.com/graph/api/resources/intune-deviceconfigv2-devicemanagementconfigurationchoicesettingvalue?view=graph-rest-beta). The specification requires recursive inspection at `/ztspecs/specs/secops/41049.md#L150-L151`.

A schema-valid mocked primary child under a choice collection produced Investigate instead of Pass. There is no evidence that the current `DisableVirusUI` setting is emitted in that subtype, so this remains a forward-compatibility warning.

**Recommendation:** recurse through every Graph setting-instance subtype that can own children, including every value in `choiceSettingCollectionValue`.

src/powershell/tests/Test-Assessment.41049.ps1
```diff
@@ -0,0 +1,366 @@
...
+
+    # Determine overall outcome based on DisableVirusUI primary control
+    $primaryResults = @($evaluationResults | Where-Object { $_.IsPrimary -eq $true })
+    $passed = $false
+    $customStatus = $null
+
+    if ($primaryResults.Count -eq 0) {
+        # No DisableVirusUI control found on either surface
+        $customStatus = 'Investigate'
+        $testResultMarkdown = "⚠️ The `DisableVirusUI` state could not be determined on either the Settings Catalog or legacy endpoint protection surface; verify configuration in the Intune portal."
```
alexandair: **No-primary Investigate results discard collected detail rows**

When no primary row exists, the message at `Test-Assessment.41049.ps1#L284-L287` contains no `%TestResult%` token. Report generation only inserts the table by replacing that token at `Test-Assessment.41049.ps1#L314-L349`.

With a legacy profile containing only contextual controls, the test correctly emitted Investigate but omitted the collected table entirely. The specification requires the Investigate details to identify queried surfaces and the reason for inconclusiveness at `/ztspecs/specs/secops/41049.md#L128-L132`, and defines a normalized output table at `/ztspecs/specs/secops/41049.md#L160-L166`.

**Recommendation:** include `%TestResult%` in the no-primary Investigate message whenever rows were collected, and explicitly report which surfaces completed or failed.

src/powershell/tests/Test-Assessment.41049.ps1
```diff
@@ -0,0 +1,366 @@
...
+    $legacyEppError = $null
+    $investigateParams = @{
+        TestId       = '41049'
+        Title        = 'The Microsoft Defender Antivirus user interface is hidden from end users'
+        Status       = $false
+        Result       = '⚠️ Unable to determine DisableVirusUI status due to an API or access error. Re-run the assessment after verifying Intune licensing, DeviceManagementConfiguration.Read.All consent, and Microsoft Graph access.'
+        CustomStatus = 'Investigate'
+    }
+
+    # Q1: List Settings Catalog configuration policies from Intune and check for Defender UI suppression settings.
```
alexandair, line 47: **Q1 processes non-MDM Settings Catalog policies**

The implementation lists all Settings Catalog policies and requests details for each at `/Test-Assessment.41049.ps1#L54-L70`. The specification limits this enumeration to policies whose `technologies` include `mdm` at `/ztspecs/specs/secops/41049.md#L134-L136`.

This adds avoidable requests and exposes the test to errors from irrelevant policies; the immediate-return behavior in critical issue "Per-policy errors return before `Fail > Investigate > Pass` is applied" amplifies that risk.

**Recommendation:** filter Q1 policies to the specified MDM technology before requesting settings and assignments.

alexandair (changes requested):

alexandair (approved):
LGTM

PR #1539, opened 2026-08-31
SecOps - 41119 - Local modification of Microsoft Defender Antivirus policy is blocked

sandeepjha000:
[Issue-1143](https://github.com/microsoft/ztspecs/issues/1143)
[Spec-41119](https://github.com/microsoft/ztspecs/blob/main/specs/secops/41119.md)


src/powershell/tests/Test-Assessment.41119.ps1
```diff
@@ -0,0 +1,332 @@
...
+            $details = 'Disable local admin merge is set to no.'
+        }
+        elseif ($assigned) {
+            $status = 'Pass'
+            $details = 'Disable local admin merge is set to yes on an assigned policy.'
+        }
+        else {
+            $status = 'Fail'
+            $details = 'Disable local admin merge is set to yes but the policy is not assigned.'
+        }
```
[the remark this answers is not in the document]
sandeepjha000: This is by design per the [spec](https://github.com/microsoft/ztspecs/blob/main/specs/secops/41119.md), so I'm keeping the current behavior.

src/powershell/tests/Test-Assessment.41119.ps1
```diff
@@ -0,0 +1,350 @@
...
+        if ([string]::IsNullOrWhiteSpace($templateFamily)) { $templateFamily = 'none' }
+
+        # PolicySummaryBlade deep link requires technology, templateId and platformName segments.
+        $technologies = [string]$policy.technologies
+        $templateId = [string]$policy.templateReference.templateId
+        $platforms = [string]$policy.platforms
+
+        # Assignment comes from the $expand=assignments projection on the list query;
+        # isAssigned is not returned by default, so the inline assignment array is the signal.
+        $assignmentCount = @($policy.assignments | Where-Object { $null -ne $_ }).Count
```
alexandair, line 144: **Conflicting or missing assignment evidence can produce a false Pass or Fail**

`Test-Assessment.41119.ps1#L143-L147` collapses assignment evidence into a Boolean with an OR:

```powershell
$assignmentCount = @($policy.assignments | Where-Object { $null -ne $_ }).Count
$isAssigned = $null
if ($null -ne $policy.isAssigned) { $isAssigned = [bool]$policy.isAssigned }
$assigned = ($isAssigned -eq $true) -or ($assignmentCount -ge 1)
```

The specification says an uninterpretable assignment state is **Investigate**, but this expression cannot produce an unknown state:

- `isAssigned = false` with one or more inline assignments becomes assigned and can Pass.
- `isAssigned = true` with an empty assignments array also becomes assigned and can Pass.
- Missing `isAssigned` and missing/malformed `assignments` become unassigned and can Fail.

These are contradictory or incomplete responses, not definite compliance outcomes. A partial or malformed Graph response could therefore produce a false Pass or false Fail.

**Required fix:** classify assignment state separately as Assigned, Unassigned, or Unknown. Use Pass only when the returned signals establish assignment, Fail only when they establish no assignment, and Investigate when the signals conflict or required assignment data is absent. Add focused tests for consistent assigned/unassigned data, conflicting signals, and missing assignment properties.

src/powershell/tests/Test-Assessment.41119.ps1
```diff
@@ -0,0 +1,350 @@
...
+                SettingDefinitionId = 'N/A'
+                RawSettingValue     = 'N/A'
+                NormalizedState     = 'Unknown'
+                Details             = "The policy settings could not be read ($statusText)."
+                Status              = 'Investigate'
+            }
+            continue
+        }
+
+        # Identify the control by exact, case-insensitive settingDefinitionId equality.
```
alexandair, line 198: **Only the first matching setting instance is evaluated**

`Test-Assessment.41119.ps1#L184-L187` discards every matching instance after the first:

```powershell
$matchingInstance = @($allInstances | Where-Object {
    [string]$_.settingDefinitionId -ieq $controlId
}) | Select-Object -First 1
```

The specification requires classification of each retrieved `DisableLocalAdminMerge` instance. If Graph returns duplicate instances, for example one enabled and one disabled in nested data, traversal order decides the result. An enabled first instance can hide a disabled later instance and incorrectly Pass the policy.

**Required fix:** retain all matching instances and evaluate all values. Apply Fail over Investigate over Pass within the policy, or classify duplicate/conflicting values as Investigate if duplicates are considered malformed. Add a regression test with conflicting duplicate instances.

src/powershell/tests/Test-Assessment.41119.ps1
```diff
@@ -0,0 +1,350 @@
...
+
+    if ($isTruncated) {
+        $tableRows += "| ... | ... | ... | ... | ... | ... | ... | ... | ... |"
+    }
+
+    if ($tableRows.Count -gt 0) {
+        $countLine = ''
+        if ($isTruncated) {
+            $countLine = "Total policies: $totalCount (showing first 10)`n`n"
+        }
```
alexandair, line 347: **The truncation row does not show omitted and total counts**

The specification requires the appended `...` row to contain omitted and total row counts. `Test-Assessment.41119.ps1#L316-L323` adds a row containing only ellipses. The separate line reports the total and first ten, but the required omitted count is absent.

Calculate `$omittedCount = $totalCount - 10` and put wording such as `... 7 omitted; 17 total ...` in the appended row.

alexandair (changes requested):

src/powershell/tests/Test-Assessment.41119.ps1
sandeepjha000: resolved in commit: [693b2a5](https://github.com/microsoft/zerotrustassessment/pull/1539/commits/693b2a5b76dbecf6608bfdea558afc178828d031)

src/powershell/tests/Test-Assessment.41119.ps1
sandeepjha000: resolved in commit: [693b2a5](https://github.com/microsoft/zerotrustassessment/pull/1539/commits/693b2a5b76dbecf6608bfdea558afc178828d031)

src/powershell/tests/Test-Assessment.41119.ps1
sandeepjha000: resolved in commit: [693b2a5](https://github.com/microsoft/zerotrustassessment/pull/1539/commits/693b2a5b76dbecf6608bfdea558afc178828d031)

src/powershell/tests/Test-Assessment.41119.ps1
```diff
@@ -137,13 +137,27 @@ function Test-Assessment-41119 {
         $templateId = [string]$policy.templateReference.templateId
         $platforms = [string]$policy.platforms
 
-        # Assignment comes from the $expand=assignments projection on the list query;
-        # isAssigned is not returned by default, so the inline assignment array is the signal.
+        # Assignment comes from the $expand=assignments projection on the list query.
+        # Classify as Assigned / Unassigned / Unknown so conflicting or absent signals surface as
+        # Investigate rather than a false Pass/Fail (isAssigned is not returned by default).
+        $assignmentsPresent = $null -ne $policy.PSObject.Properties['assignments']
```
alexandair, line 143: **A null assignments value is treated as definitely unassigned**

The new classifier checks whether the `assignments` property exists, but it does not check whether the property contains a usable collection:

`Test-Assessment.41119.ps1#L143-L157`

```powershell
$assignmentsPresent = $null -ne $policy.PSObject.Properties['assignments']
$assignmentCount = @($policy.assignments | Where-Object { $null -ne $_ }).Count
...
elseif ($assignmentCount -ge 1) {
    $assignmentState = 'Assigned'
}
else {
    $assignmentState = 'Unassigned'
}
```

When Graph returns an `assignments` property whose value is `$null`, `$assignmentsPresent` is true and `$assignmentCount` is zero. The code therefore sets the state to `Unassigned`. An enabled setting then becomes a definite Fail in the new per-instance logic at `Test-Assessment.41119.ps1#L239-L245`.

A null expanded collection is not the same as a valid empty array. It is missing or unusable assignment evidence, so the specification requires Investigate rather than Fail.

I confirmed the behavior with the exact classifier:

| Input | Current state |
|---|---|
| `assignments` property absent | Unknown |
| `assignments = $null` | Unassigned |
| `assignments = @()` and `isAssigned = $false` | Unassigned |
| One assignment and `isAssigned = $false` | Unknown |

**Fix:** inspect both the property and its value. Treat an absent or null value as Unknown, while preserving an actual empty collection as Unassigned. Add a regression test for `assignments = $null` with an enabled control and verify the result is Investigate.

alexandair (changes requested):

